PR #11, opened 2017-12-22
Add better error handling

mattolson:
## What? Why?
* Add error handling around restoring precompiled templates, rendering templates, compiling templates, looking up templates
* Update addTemplates so it can handle a mix of precompiled templates and raw template strings
* Better detection of precompiled templates before doing a potentially dangerous eval.
* Add logger and ability for caller to set a logger

## How was it tested?
`npm test`

----

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const _ = require('lodash');
+const Logger = require('./lib/logger');
 const HandlebarsV3 = require('handlebars');
 const HandlebarsV4 = require('@bigcommerce/handlebars-v4');
 const helpers = require('./helpers');
@@ -34,6 +35,7 @@ class HandlebarsRenderer {
         this._translator = null;
         this._decorators = [];
         this._contentRegions = {};
+        this._logger = new Logger();
 
         // Build global context for helpers
         this.helperContext = {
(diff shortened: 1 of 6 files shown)


mattolson:
@bigcommerce/storefront-team

lib/logger.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+class Logger {
+    constructor() {}
+
```
tabayomi: All these methods output error to the console? Since it's server side, I guess I'm curious as to why as opposed to a file?

mattolson: The default logger is really simple and just outputs to the console, but I added an accessor so that an external logger can be passed in. We will supply paper-handlebars with a more robust logger that does what we want for our infrastructure.

index.js
```diff
@@ -68,23 +88,55 @@ class HandlebarsRenderer {
...
      */
     addTemplates(templates) {
-        _.each(templates, (precompiled, path) => {
-            // Take precompiled template and convert it into something that can
-            // be directly executed. Then, add it to the list of templates & partials.
-            if (typeof this.handlebars.partials[path] === 'undefined') {
-                let template;
-                eval(`template = ${precompiled};`); // evil, but necessary :/
-                this.handlebars.registerPartial(path, this.handlebars.template(template));
+        _.each(templates, (template, path) => {
```
tabayomi: If the only reason `lodash` is included here is for looping, consider using the native `es6` implementations. It's probably microseconds savings, but also, maybe less `mb`s? 🍹

mattolson: We won't be able to get rid of lodash entirely because many of the helpers depend on it.

PR #15, opened 2018-01-31
STRF-4420: Fix scope of call to getSiteSettings

mattolson:
## What? Why?
In the `cdn` and `stylesheet` helpers, we were making the call to `getThemeSettings` and `getSiteSettings` at the wrong time, so not getting the latest value if they were updated after construction.

## How was it tested?
Unit tests added

----

### DIFF
--- a/helpers/lib/cdnify.js
+++ b/helpers/lib/cdnify.js
@@ -2,20 +2,20 @@
 
 // Return a function that can be used to translate paths to cdn paths
 module.exports = globals => {
-    const siteSettings = globals.getSiteSettings();
-    const themeSettings = globals.getThemeSettings();
-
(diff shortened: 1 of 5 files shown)


helpers/lib/cdnify.js
```diff
@@ -2,20 +2,20 @@
...
-    const editSessionId = siteSettings.theme_session_id;
-    const cdnSettings = themeSettings.cdn;
-
     /**
      * Add CDN base url to the relative path
      * @param  {String} path     Relative path
      * @return {String}          Url cdn
      */
     return function(path) {
+        const siteSettings = globals.getSiteSettings();
```
mjschock: interesting - didn't think about it before but it looks like this means `const` uses block scoping

PR #17, opened 2018-05-31
STRF-4917: Port new helpers from paper 2.0.8

mattolson:
## What? Why?
* Port helper updates from https://github.com/bigcommerce/paper/pull/133/files
* Update version to 3.0.2

## How was it tested?
* `npm test`
* Updated local paper to use this, then updated local storefront renderer to use local paper, then confirmed that `{{getFontCollection}}` still worked. Then updated the test client script to render `{{resourceHints}}` and `{{getFontLoaderConfig}}` and confirmed that they work.

----


spec/helpers/resourceHints.js
```diff
@@ -0,0 +1,30 @@
...
+            'test6-font': 'Google_Volkron',
+            'test7-font': 'Google_Droid_400,700',
+            'test8-font': 'Google_Crimson+Text_400,700_sans',
+            'random-property': 'not a font'
+        };
+
+        const template = "{{resourceHints}}";
+
+        expect(renderString(template, {}, {}, themeSettings))
+            .to.be.equal('<link rel="dns-prefetch preconnect" href="//ajax.googleapis.com" crossorigin><link rel="dns-prefetch preconnect" href="//fonts.googleapis.com" crossorigin><link rel="dns-prefetch preconnect" href="//fonts.gstatic.com" crossorigin>');
```
tpietsch: is this dom order always promsied? Like can the links be added in a different order - thinking intermittent test failures

mattolson: Yeah, it is due to the way the implementation is written.

PR #19, opened 2018-08-06
added under if helpers comparison of number string

bc-zoharmuzafi:
## What? Why?
Under the "if" helper we didn't have a case for number strings that convert string that are numbers and compare them.This pr will add "gtnum" which takes an input of string -> convert them to integers and check if left grater than right. 

## How was it tested?
specs and manually.


bookernath:
FYI @bigcommerce/dev-docs

PR #21, opened 2018-10-29
Handle case in cdn helper for img/icon-sprite.svg

bc-zoharmuzafi:
## What? Why?
Currently we are generating cdn link for img/icon-sprite.svg with double slash (//img/icon-sprite.svg). This fix handle this case and avoid double slash.

## How was it tested?
Unitest and manually.

----

### DIFF
--- a/helpers/lib/cdnify.js
+++ b/helpers/lib/cdnify.js
@@ -60,16 +60,20 @@ module.exports = globals => {
             return path;
         }
 
-        if (path[0] !== '/') {
-            path = '/' + path;
-        }
 
         if (!versionId) {
+            if (path[0] !== '/') {
+                path = '/' + path;
+            }
             return path;
         }
 
-        if (path.match(/^\/assets\//)) {
-            path = path.substr(8, path.length);
+        if (path[0] === '/') {
+            path = path.slice(1, path.length);
+        }
+
+        if (path.match(/^assets\//)) {
+            path = path.substr(7, path.length);
         }
 
         if (editSessionId) {
--- a/spec/helpers/cdn.js
+++ b/spec/helpers/cdn.js
@@ -209,4 +209,13 @@ describe('cdn helper', function () {
             },
         ], done);
     });
+
+    it('should avoid double slash in path', function (done) {
+        runTestCases([
+            {
+                input: '{{cdn "img/icon-sprite.svg"}}',
+                output: 'https://cdn.bcapp/3dsf74g/stencil/123/img/icon-sprite.svg',
+            }
+        ], done);
+    });
 });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bigcommerce/stencil-paper-handlebars",
-  "version": "4.0.0",
+  "version": "4.0.1",
   "description": "A paper plugin to render pages using Handlebars.js",
   "main": "index.js",
   "author": "Bigcommerce",


helpers/lib/cdnify.js
```diff
@@ -60,7 +60,7 @@ module.exports = globals => {
             return path;
         }
 
-        if (path[0] !== '/') {
+        if ((path !== 'img/icon-sprite.svg') && (path[0] !== '/')) {
```
mattolson: I wouldn't want to introduce any hardcoded paths like this. Let's dig deeper to find the underlying root cause.

bc-zoharmuzafi: @mattolson I pushed a change for that

helpers/lib/cdnify.js
```diff
@@ -76,6 +78,8 @@ module.exports = globals => {
             return [cdnUrl, 'stencil', versionId, 'e', editSessionId, path].join('/');
         }
 
-        return [cdnUrl, 'stencil', versionId, path].join('/');
+        urlElements = [cdnUrl, 'stencil', versionId, path].map((element, idx) => { return (element[0] === '/' || idx === 0) ?  element : "/" + element });
+        
+        return urlElements.join("");
```
mattolson: This doesn't solve all return paths from this function.

I think the solution to this issue is much easier. Before calling `join('/')`, we must ensure that `path` does not contain a leading `/`. You can see that we are explicitly adding that above. Instead of adding the leading `/`, we should remove it.

bc-zoharmuzafi: @mattolson  This leading / is required to the check of the check of the assets: https://github.com/bigcommerce/paper-handlebars/blob/master/helpers/lib/cdnify.js#L71
If we not adding / in places that this one is exist the returned path is correct

helpers/lib/cdnify.js
```diff
@@ -72,6 +72,10 @@ module.exports = globals => {
             path = path.substr(8, path.length);
         }
 
+        if (path[0] === '/') {
+            path = path.slice(1, path.length);
+        }
+
```
mattolson: How about something like this?

```
        if (!versionId) {
             if (path[0] !== '/') {
                 path = '/' + path;
             }
     
             return path;
         }

         if (path[0] === '/') {
            path = path.slice(1, path.length);
        }

         if (path.match(/^assets\//)) {
            path = path.substr(8, path.length);
        }
         
         if (editSessionId) {
            return [cdnUrl, 'stencil', versionId, 'e', editSessionId, path].join('/');
        }
 
         return [cdnUrl, 'stencil', versionId, path].join('/');
```

bc-zoharmuzafi: @mattolson I pushed a new pr.

spec/helpers/cdn.js
```diff
@@ -209,4 +209,13 @@ describe('cdn helper', function () {
             },
         ], done);
     });
+
+    it('should handle case for img/icon-sprite.svg and remove / at the beggining of path before join', function (done) {
```
mattolson: Please change this description to something simpler and more generic, like `should not result in empty path segments`

mattolson: Also please squash your commits.

helpers/lib/cdnify.js
```diff
@@ -60,16 +60,20 @@ module.exports = globals => {
             return path;
         }
 
-        if (path[0] !== '/') {
-            path = '/' + path;
-        }
 
         if (!versionId) {
+            if (path[0] !== '/') {
```
junedkazi: Instead of nesting can we just return path early if version exists ?

mattolson: This logic is already convoluted enough, I think this is the best we can do right now without a much bigger refactor.

PR #27, opened 2019-01-30
trimmed . and / in cdn helper for paths which are not full path

bc-zoharmuzafi:
## What? Why?
Currently, cdn helper is not handling relative path and not trim . or / from the begging of the path before creating cdn url. That cause paths such as ../example.com ,  ./example.com ,  /example.com,  ../../example.com, etc to generate wrong cdn url which leads to a wrong request by the browser (return 404)

## How was it tested?
Unitests

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 4.0.4
+- Added trimming of all '.' and '/' from begginig of cdn path (not full paths)
+
 ## 4.0.3
 - Fix resourceHints to use https for font providers
 
--- a/helpers/lib/cdnify.js
+++ b/helpers/lib/cdnify.js
@@ -31,7 +31,13 @@ module.exports = globals => {
         }
 
         if (['.', '/'].includes(path[0])) {
-            path = path.replace(/^[^a-z\d]*/gi, '');
+            for (var i = 0; i < path.length; i++) {
+                if (['.', '/'].includes(path[0])) {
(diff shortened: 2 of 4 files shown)


helpers/lib/cdnify.js
```diff
@@ -30,6 +30,10 @@ module.exports = globals => {
             return path;
         }
 
+        if (['.', '/'].includes(path[0])) {
+            path = path.replace(/^[^a-z\d]*/gi, '');
```
mattolson: Let's just remove `.` and `/` -- make this a strict selection rather than a negation.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 4.0.4
+- Added trimmimg of all . and / chars from begginig of cdn path (not full paths)
```
mattolson: Typo

helpers/lib/cdnify.js
```diff
@@ -30,6 +30,16 @@ module.exports = globals => {
             return path;
         }
 
+        if (['.', '/'].includes(path[0])) {
```
mattolson: In general, modifying a string that you are iterating through is a bad practice. A much cleaner implementation would be:

```
path = path.replace(/^[\.\/]+/, '');
```

This is similar to your previous solution. My comment at that time was to make the characters you are selecting for more explicit -- i.e. to use `[\.\/]` instead of `[^a-zA-Z0-9]`.

PR #29, opened 2019-02-14
STRF-6040 Fix regex performance to match precompiled templates.

junedkazi:
## What? Why?
Fix regex performance to match precompiled templates.
`.*` at the beginning is forcing it to match all characters and for a big precompiled template file it takes over 10 sec.

## How was it tested?
The regex was matching every 

cc @bigcommerce/storefront-team @bigcommerce/platform-engineering

### DIFF
--- a/index.js
+++ b/index.js
@@ -190,7 +190,7 @@ class HandlebarsRenderer {
         // be a string representation of an object containing a `main`
         // function and a `compiler` array. We do this because the next
         // step is a potentially dangerous eval.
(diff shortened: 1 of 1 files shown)


mattolson:
picard-facepalm

zvuki:
How fast is the new regex?
Do we need to keep a regex?

junedkazi:
@zvuki before 9000ms and after 1ms.

bookernath:
Crazy.

PR #35, opened 2019-05-23
STRF-6687 Change default behavior of {{getFontsCollection}} to use font-display: swap; for Google Fonts, allow configuration

bc-williamkwon:
## What? Why?
Change default behavior of {{getFontsCollection}} to use font-display: swap; for Google Fonts, allow configuration
## How was it tested?
  
----
Unit tested can run test suite to check.

### DIFF
--- a/helpers/getFontsCollection.js
+++ b/helpers/getFontsCollection.js
@@ -4,7 +4,9 @@ const getFonts = require('./lib/fonts');
 
 const factory = globals => {
     return function() {
-        return getFonts('linkElements', globals.getThemeSettings(), globals.handlebars);
+        const options = arguments.length === 1 ? arguments[arguments.length - 1] : null;
+        const fontDisplay = options && options.hash['font-display'] ? options.hash['font-display'] : null;
+        return getFonts('linkElements', globals.getThemeSettings(), globals.handlebars, {fontDisplay});
     };
 };
 
--- a/helpers/lib/fonts.js
+++ b/helpers/lib/fonts.js
@@ -51,8 +51,10 @@ const fontProviders = {
             return collection;
         },
 
-        buildLink: function(fonts) {
-            return '<link href="https://fonts.googleapis.com/css?family=' + fonts.join('|') + '" rel="stylesheet">';
+        buildLink: function(fonts, fontDisplay) {
+            const displayTypes = ['auto', 'block', 'swap', 'fallback', 'optional'];
+            fontDisplay = displayTypes.includes(fontDisplay) ? fontDisplay : 'swap';
+            return `<link href="https://fonts.googleapis.com/css?family=${fonts.join('|')}&display=${fontDisplay}" rel="stylesheet">`;
         },
 
         buildFontLoaderConfig: function(fonts) {
@@ -78,10 +80,11 @@ const fontProviders = {
  *   object that can be used to configure Web Font Loader.
  * @param {Object} themeSettings Object containing theme settings.
  * @param {Object} handlebars The handlebars instance.
+ * @param {Object} options an optional object for additional configuration details
  * @returns {Object.<string, Array>|string}
  */
-module.exports = function(format, themeSettings, handlebars) {
-    // Collect font strings from theme settings
+module.exports = function(format, themeSettings, handlebars, options) {
+    
     const collectedFonts = {};
     _.each(themeSettings, function(value, key) {
         if (!fontKeyFormat.test(key)) {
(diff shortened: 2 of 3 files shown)


helpers/lib/fonts.js
```diff
@@ -51,8 +51,8 @@ const fontProviders = {
             return collection;
         },
 
-        buildLink: function(fonts) {
-            return '<link href="https://fonts.googleapis.com/css?family=' + fonts.join('|') + '" rel="stylesheet">';
+        buildLink: function(fonts, fontDisplay) {
+            return `<link href="https://fonts.googleapis.com/css?family=${fonts.join('|')}&display=${fontDisplay ? fontDisplay : 'swap'}" rel="stylesheet">`;
```
davidchin: You might want to do a check to make sure the value is valid, i.e.: `auto | block | swap | fallback | optional`. 🍺

helpers/getFontsCollection.js
```diff
@@ -3,8 +3,8 @@
 const getFonts = require('./lib/fonts');
 
 const factory = globals => {
-    return function() {
-        return getFonts('linkElements', globals.getThemeSettings(), globals.handlebars);
+    return function(options) {
```
davidchin: Do we need to document the parameter? Or do you guys usually just document it separately on DevDocs? Regardless, it might be a good practice to have a docblock showing how it could be applied (i.e.: `@example tag`). 🍺

helpers/lib/fonts.js
bc-williamkwon: Thanks, added that check and added unit tests for it. Made it go to default if invalid.

helpers/getFontsCollection.js
bc-williamkwon: Changed it to the arguments pattern that I saw other people use.

davidchin: Right, I think it might be better if you keep the parameter name because it's easier to read. I just want a comment showing that it accepts 'font-display' as a valid optional parameter. At the moment, it's not immediately obvious unless you go through the code. 😄

helpers/lib/fonts.js
```diff
@@ -51,8 +51,10 @@ const fontProviders = {
             return collection;
         },
 
-        buildLink: function(fonts) {
-            return '<link href="https://fonts.googleapis.com/css?family=' + fonts.join('|') + '" rel="stylesheet">';
+        buildLink: function(fonts, fontDisplay) {
+            const displayTypes = ['auto', 'block', 'swap', 'fallback', 'optional'];
+            fontDisplay = displayTypes.includes(fontDisplay) ? fontDisplay : null;
```
mattolson: Would be simpler if you just made `swap` the default value here. Then you wouldn't need an additional conditional in the string construction on the next line.

```
fontDisplay = displayTypes.includes(fontDisplay) ? fontDisplay : 'swap';
return `<link href="https://fonts.googleapis.com/css?family=${fonts.join('|')}&display=${fontDisplay}" rel="stylesheet">`;
```

helpers/lib/fonts.js
```diff
@@ -110,8 +112,10 @@ module.exports = function(format, themeSettings, handlebars) {
     // Format output based on requested format
     switch(format) {
     case 'linkElements':
+        const options = arguments.length > 3 ? arguments[arguments.length - 1] : null;
+        const fontDisplay = options && options.hash['font-display'] ? options.hash['font-display'] : null;
```
mattolson: Please move this use of `arguments` up to the helper level (in `getFontsCollection`). Since this is library code and not an external interface with theme code, we have more flexibility in changing the interface and don't need to rely on this hack. Instead, it would be better to add an `options` object as an optional fourth parameter to this function, and then create these options in `getFontCollection` as needed.

bc-williamkwon: Removed arguments line.

helpers/lib/fonts.js
bc-williamkwon: Refactored logic.

helpers/lib/fonts.js
bc-williamkwon: Thanks for the suggestions, I can't believe I didn't think of the first one lol.

helpers/lib/fonts.js
```diff
@@ -110,8 +113,9 @@ module.exports = function(format, themeSettings, handlebars) {
     // Format output based on requested format
     switch(format) {
     case 'linkElements':
+        const fontDisplay = options && options.hash['font-display'] ? options.hash['font-display'] : null;
```
mattolson: This is much better, but it's still kind of awkward that this library code needs to know about how Handlebars handles options (i.e. that some are plain variables and some are in the `hash` object. I think the `getFontsCollection` helper should translate those concerns and then pass in a normal options object to this library code. This would also act as documentation in the helper as to what options are supported (something David Chin asked for as well).

bc-williamkwon: ok I changed it again.

mattolson: Perfect

PR #40, opened 2019-05-29
Refactor helpers to reduce argument usage

bc-williamkwon:
## What? Why?

https://jira.bigcommerce.com/browse/STRF-6707

CPU profiling indicates that using arguments is slow. We use this pattern quite a bit in our handlebars helpers. This pull request is to move away from that pattern.

##How was it tested?

The unit test suite tests pass after some refactoring to fit the new options pattern.  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Changelog
+## Unreleased
+- remove regex in fonts.js for better performance
 
 ## 4.0.6
 - Change default behavior of {{getFontsCollection}} to use font-display: swap for Google Fonts and allow configuration
--- a/helpers/lib/fonts.js
+++ b/helpers/lib/fonts.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const _ = require('lodash');
-const fontKeyFormat = new RegExp(/\w+(-\w*)*-font$/);
+
 const fontProviders = {
     'Google': {
         /**
@@ -87,7 +87,9 @@ module.exports = function(format, themeSettings, handlebars, options) {
     
     const collectedFonts = {};
     _.each(themeSettings, function(value, key) {
-        if (!fontKeyFormat.test(key)) {
+        //check that -font is on end of string but not start of string
+        const fontKeySuffix = '-font';
+        if (!key.endsWith(fontKeySuffix)) {
             return;
         }
 


mattolson:
@bc-williamkwon Can you provide a more details PR description showing the rationale for this change? Since this is an open source repository, and the public does not have access to our internal ticket system, it is not as clear as it could be.

bc-williamkwon:
Just did.

helpers/partial.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 const factory = globals => {
-    return function(name) {
-        const options = arguments[arguments.length - 1];
+    return function(name, options) {
+        options = options ? options : name;
```
mattolson: I think name should be required here.

helpers/compare.js
```diff
@@ -1,13 +1,13 @@
...
-    return function(lvalue, rvalue) {
-        const options = arguments[arguments.length - 1];
+    return function(lvalue, rvalue, options) {
+
         var operator;
         var operators;
         var result;
 
-        if (arguments.length < 3) {
+        if (options === undefined) {
```
mattolson: The most robust way to check for undefined is `typeof x === 'undefined'`

helpers/for.js
```diff
@@ -3,8 +3,8 @@
 const _ = require('lodash');
 
 const factory = () => {
-    return function(from, to, context) {
-        const options = arguments[arguments.length - 1];
+    return function(from, to, context, options) {
+        options =  options ? options : context ? context : to;
```
mattolson: Nested ternary operators should probably be disambiguated with parentheses. Even if Javascript knows what to do here, it's confusing for the reader.

helpers/if.js
```diff
@@ -3,8 +3,8 @@
 const _ = require('lodash');
 
 const factory = globals => {
-    return function(lvalue, operator, rvalue) {
-        const options = arguments[arguments.length - 1];
+    return function(lvalue, operator, rvalue, options) {
+        options = options ? options : rvalue ? rvalue : operator;
```
mattolson: Same thing here -- please disambiguate.

spec/helpers/block.js
```diff
@@ -37,7 +37,7 @@ describe('partial and block helpers', function () {
 
     it('should not trigger an error if block name is empty', function (done) {
         var templates = {
-            template: '{{#partial "page" "2134"}}World{{/partial}}{{> layout}}',
+            template: '{{#partial "page"}}World{{/partial}}{{> layout}}',
```
mattolson: Why did this test need to change?

spec/helpers/stylesheet.js
```diff
@@ -25,7 +25,7 @@ describe('stylesheet helper', () => {
     it('should render a link tag and all extra attributes with no cdn url', done => {
         runTestCases([
             {
-                input: '{{{stylesheet "assets/css/style.css" blah rel="something" class="myStyle"}}}',
+                input: '{{{stylesheet "assets/css/style.css" rel="something" class="myStyle"}}}',
```
mattolson: Why did this test need to change?

helpers/block.js
```diff
@@ -1,10 +1,9 @@
 'use strict';
 
 const factory = globals => {
-    return function(name) {
-        const options = arguments[arguments.length - 1];
-
+    return function(name, options) {
         /* Look for partial by name. */
+        options = options ? options : name;
```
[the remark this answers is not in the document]
bc-williamkwon: There are tests that say should not fail if partial name is empty

helpers/partial.js
bc-williamkwon: There are tests that say should not trigger an error if partial name is empty.

helpers/compare.js
bc-williamkwon: will change that.

spec/helpers/block.js
bc-williamkwon: test needed to change as more than one argument is being passed to partial helper.  We do not have access to options in helper if that is the case.

spec/helpers/stylesheet.js
bc-williamkwon: test needed to change as more than one argument is being passed to stylesheet.js helper.  We do not have access to options in helper if that is the case.

helpers/if.js
bc-williamkwon: will do.

helpers/for.js
bc-williamkwon: will do.

helpers/compare.js
```diff
@@ -1,13 +1,13 @@
...
-    return function(lvalue, rvalue) {
-        const options = arguments[arguments.length - 1];
+    return function(lvalue, rvalue, options) {
+
         var operator;
         var operators;
         var result;
 
-        if (arguments.length < 3) {
+        if (typeof options === undefined) {
```
mattolson: This should be `typeof options === 'undefined'`

mattolson: Also, I would change the error message to say "2 parameters and a block"

helpers/for.js
```diff
@@ -3,8 +3,8 @@
 const _ = require('lodash');
 
 const factory = () => {
-    return function(from, to, context) {
-        const options = arguments[arguments.length - 1];
+    return function(from, to, context, options) {
+        options =  options ? options : (context ? context : to);
```
mattolson: It's not immediately apparent how this relates to the parameter shuffling that takes place below and why we wouldn't do this at the same time.

helpers/getFontsCollection.js
```diff
@@ -3,9 +3,8 @@
 const getFonts = require('./lib/fonts');
 
 const factory = globals => {
-    return function() {
-        const options = arguments.length === 1 ? arguments[arguments.length - 1] : null;
-        const fontDisplay = options && options.hash['font-display'] ? options.hash['font-display'] : null;
+    return function(options) {
+        const fontDisplay = options.hash['font-display'];
```
mattolson: Do we need to check for the presence of options here, or is options and options.hash always defined?

helpers/if.js
```diff
@@ -3,8 +3,8 @@
 const _ = require('lodash');
 
 const factory = globals => {
-    return function(lvalue, operator, rvalue) {
-        const options = arguments[arguments.length - 1];
+    return function(lvalue, operator, rvalue, options) {
+        options = options ? options : (rvalue ? rvalue : operator);
```
mattolson: It's not immediately apparent how this relates to the parameter shuffling that takes place below and why we wouldn't do this at the same time.

helpers/partial.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 const factory = globals => {
-    return function(name) {
-        const options = arguments[arguments.length - 1];
+    return function(name, options) {
```
mattolson: Perhaps the safer thing is

```
function(...args) {
  const options = args.pop();
  const name = args[0];
}
```

Then the tests won't have to change, perhaps that test was added for backward compatibility with an unused parameter.

helpers/getFontsCollection.js
bc-williamkwon: If they aren't passing extra arguments then options and options.hash should be defined.  If nothing extra is passed options.hash is empty object.

helpers/compare.js
bc-williamkwon: The thing is that the user passes two arguments and the block is provided by Handlebars.

bc-williamkwon: Actually, I will change that now.

helpers/partial.js
bc-williamkwon: fixed for partial.js, block.js and stylesheet.js set test cases back to normal.

helpers/for.js
bc-williamkwon: added a comment

helpers/if.js
bc-williamkwon: added a comment

helpers/for.js
```diff
@@ -3,8 +3,10 @@
 const _ = require('lodash');
 
 const factory = () => {
-    return function(from, to, context) {
-        const options = arguments[arguments.length - 1];
+    return function(from, to, context, options) {
+        //function takes variable number of arguments, options is last parameter and also needed for
```
mattolson: I don't like the fact that we are using two patterns for argument shuffling in this function. There is code below that does something different, including a stricter check on the options object. I feel like you should integrate your options param into that code rather than doing this here.

helpers/if.js
```diff
@@ -3,8 +3,10 @@
 const _ = require('lodash');
 
 const factory = globals => {
-    return function(lvalue, operator, rvalue) {
-        const options = arguments[arguments.length - 1];
+    return function(lvalue, operator, rvalue, options) {
+        //function takes variable number of arguments, options is last parameter and also needed for
```
mattolson: I don't like the fact that we are using two patterns for argument shuffling in this function. There is code below that does something different, including a stricter check on the options object. I feel like you should integrate your options param into that code rather than doing this here.

bc-williamkwon: fixed.

helpers/for.js
bc-williamkwon: fixed.

bc-williamkwon: I think there was weird git behavior that lead to for and if not being changed on last rebase.

PR #41, opened 2019-06-01
STRF-6701 Reduce usage of Lodash

bc-williamkwon:
## What? Why?
 Refactored helper to use vanilla javascript methods instead.

https://jira.bigcommerce.com/browse/STRF-6701 

## How was it tested?

----
Unit tests still pass.

### DIFF
--- a/helpers/contains.js
+++ b/helpers/contains.js
@@ -1,7 +1,6 @@
 'use strict';
 
-const _ = require('lodash');
-
+const common = require('./lib/common.js');
 /**
  * Is any value included in a collection or a string?
  *
@@ -10,10 +9,14 @@ const _ = require('lodash');
  * {{#contains font_path "Roboto"}} ... {{/contains}}
  */
 const factory = () => {
-    return function() {
-        var args = Array.prototype.slice.call(arguments, 0, -1),
-            options = _.last(arguments),
-            contained = _.contains.apply(_, args);
+    return function(container, value, options) {
+        let contained;
+
+        if (Array.isArray(container) || common.isString(container)) {
+            contained = container.includes(value);
+        } else if (common.isObject(container)) {
+            contained = Object.values(container).includes(value);
+        }
 
         // Yield block if true
         if (contained) {
--- a/helpers/lib/fonts.js
+++ b/helpers/lib/fonts.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
-
 const fontProviders = {
     'Google': {
         /**
@@ -18,35 +16,35 @@ const fontProviders = {
          * @returns {string}
          */
         parser: function(fonts) {
-            var collection = [],
-                familyHash = {};
+            const collection = [];
+            const familyHash = {};
 
-            _.each(fonts, function fontsIterator(font) {
-                var split = font.split('_'),
-                    familyKey = split[1],  // Eg: Open+Sans
-                    weights = split[2];    // Eg: 400,700italic
+            for (let i = 0; i < fonts.length; i++) {
+                const split = fonts[i].split('_');
+                const familyKey = split[1];
+                let weights = split[2];
 
-                if (_.isEmpty(familyKey)) {
-                    return;
+                if (!familyKey || !familyKey.length) {
+                   continue;
                 }
-
-                if (_.isUndefined(weights)) {
+                if (typeof weights === 'undefined') {
                     weights = '';
                 }
-
-                if (!_.isArray(familyHash[familyKey])) {
+                if (!Array.isArray(familyHash[familyKey])) {
                     familyHash[familyKey] = [];
                 }
-
                 weights = weights.split(',');
+                familyHash[familyKey].push(...weights);
+            }
 
-                familyHash[familyKey].push(weights);
-                familyHash[familyKey] = _.uniq(_.flatten(familyHash[familyKey]));
-            });
+            const familyHashKeys = Object.keys(familyHash);
 
-            _.each(familyHash, function fontHashIterator(weights, family) {
-                collection.push(family + ':' + weights.join(','));
-            });
+            for (let i = 0; i < familyHashKeys.length; i++) {
+                const family = familyHashKeys[i];
+                const weights = familyHash[family];
+                const uniqueWeights = Array.from(new Set(weights));
+                collection.push(`${family}:${uniqueWeights.join(',')}`);
+            }
 
             return collection;
         },
@@ -61,10 +59,16 @@ const fontProviders = {
             function replaceSpaces(font) {
                 return font.split('+').join(' ');
             }
-
+            function familyBuilder(fonts) {
+                const fontsCopy = fonts.slice();
+                for (let i = 0; i < fontsCopy.length; i++) {
+                    fontsCopy[i] = replaceSpaces(fontsCopy[i]);
+                }
+                return fontsCopy;
+            }
             return {
                 google: {
-                    families: _.map(fonts, replaceSpaces),
+                    families: familyBuilder(fonts),
                 }
             };
         },
@@ -84,20 +88,24 @@ const fontProviders = {
  * @returns {Object.<string, Array>|string}
  */
 module.exports = function(format, themeSettings, handlebars, options) {
-    
+
     const collectedFonts = {};
-    _.each(themeSettings, function(value, key) {
+    const themeSettingsKeys = Object.keys(themeSettings);
+
+    for (let i = 0; i < themeSettingsKeys.length; i++) {
+        const key = themeSettingsKeys[i];
+        const value = themeSettings[key];
         //check that -font is on end of string but not start of string
         const fontKeySuffix = '-font';
         if (!key.endsWith(fontKeySuffix)) {
-            return;
+            continue;
         }
 
         const splits = value.split('_');
         const provider = splits[0];
 
         if (typeof fontProviders[provider] === 'undefined') {
-            return;
+            continue;
         }
 
         if (typeof collectedFonts[provider] === 'undefined') {
@@ -105,33 +113,51 @@ module.exports = function(format, themeSettings, handlebars, options) {
         }
 
         collectedFonts[provider].push(value);
-    });
+    }
 
     // Parse font strings based on provider
-    const parsedFonts = _.mapValues(collectedFonts, function(value, key) {
-        return fontProviders[key].parser(value);
-    });
+    const parsedFonts = {};
+    const collectedFontsKeys = Object.keys(collectedFonts);
 
+    for (let i = 0; i < collectedFontsKeys.length; i++) {
+        const font = collectedFontsKeys[i];
+        parsedFonts[font] = fontProviders[font].parser(collectedFonts[font]);
+    }
+    const parsedFontsKeys = Object.keys(parsedFonts);
     // Format output based on requested format
     switch(format) {
     case 'linkElements':
+        const formattedFonts = {};
+
+        for (let i = 0; i < parsedFontsKeys.length; i++) {
+            const font = parsedFontsKeys[i];
+            formattedFonts[font] = fontProviders[font].buildLink(parsedFonts[font], options.fontDisplay);
+        }
+        const formattedFontsOutput = [];
+        const formattedFontsKeys = Object.keys(formattedFonts);
         
-        const formattedFonts = _.mapValues(parsedFonts, function(value, key) {
-            return fontProviders[key].buildLink(value, options.fontDisplay);
-        });
-        return new handlebars.SafeString(_.values(formattedFonts).join(''));
+        for (let i = 0; i < formattedFontsKeys.length; i++) {
+            formattedFontsOutput.push(formattedFonts[formattedFontsKeys[i]]);
+        }
+        return new handlebars.SafeString(formattedFontsOutput.join(''));
 
     case 'webFontLoaderConfig':
         // Build configs
-        const configs = _.mapValues(parsedFonts, function(value, key) {
-            return fontProviders[key].buildFontLoaderConfig(value);
-        });
+        const configs = {};
+
+        for (let i = 0; i < parsedFontsKeys.length; i++) {
+            const font = parsedFontsKeys[i];
+            configs[font] = fontProviders[font].buildFontLoaderConfig(parsedFonts[font]);
+        }
 
         // Merge them
         const fontLoaderConfig = {};
-        _.each(configs, function(config) {
-            return Object.assign(fontLoaderConfig, config);
-        });
+        const configValues = Object.values(configs);
+
+        for (let i = 0; i < configValues.length; i++) {
+            Object.assign(fontLoaderConfig, configValues[i]);
+        }
+
         return fontLoaderConfig;
 
     case 'providerLists':
--- a/helpers/pluck.js
+++ b/helpers/pluck.js
@@ -1,11 +1,23 @@
 'use strict';
 
-var _ = require('lodash');
+const common = require('./lib/common.js');
 
 const factory = () => {
     return function(collection, path) {
-        return _.pluck(collection, path);
-    };
+        const arr = [];
+
+        if (Array.isArray(collection)) {
+            for (let i = 0; i < collection.length; i++) {
+                arr.push(collection[i][path]);
+            } 
+        } else if (common.isObject(collection)) {
+            const collectionKeys = Object.keys(collection);
+            for (let i = 0; i < collectionKeys.length; i++) {
+                arr.push(collection[collectionKeys[i]][path]);
+            }
+        }
+        return arr;
+    }
 };
 
 module.exports = [{
--- a/helpers/resourceHints.js
+++ b/helpers/resourceHints.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const getFonts = require('./lib/fonts');
 
 const fontResources = {
@@ -16,24 +15,21 @@ const factory = globals => {
             return `<link rel="dns-prefetch preconnect" href="${host}" crossorigin>`;
         }
 
-        var hosts = [];
-
         // Add cdn
         const siteSettings = globals.getSiteSettings();
-        const cdnUrl = siteSettings['cdn_url'] || '';
-        if (cdnUrl != '') {
-            hosts.push(cdnUrl);
-        }
+        const hosts = siteSettings['cdn_url'] ?  [siteSettings['cdn_url']] : [];
 
         // Add font providers
-        const fontProviders = _.keys(getFonts('providerLists', globals.getThemeSettings(), globals.handlebars));
-        _.each(fontProviders, function(provider) {
-            if (typeof fontResources[provider] !== 'undefined') {
-                hosts = hosts.concat(fontResources[provider]);
-            }
-        });
+        const providers = Object.keys(getFonts('providerLists', globals.getThemeSettings(), globals.handlebars));
 
-        return new globals.handlebars.SafeString(_.map(hosts, format).join(''));
+        for (let i = 0; i < providers.length; i++) {
+            if (typeof fontResources[providers[i]] !== 'undefined') {
+                for (let j = 0; j < fontResources[providers[i]].length; j++) {
+                    hosts.push(format(fontResources[providers[i]][j]));
+                }
+            }
+        }
+        return new globals.handlebars.SafeString(hosts.join(''));
     };
 };
 


helpers/all.js
```diff
@@ -21,20 +19,15 @@ const factory = () => {
...
         // Check if all the arguments are valid / truthy
-        result = _.all(args, function (arg) {
-            if (_.isArray(arg)) {
+        result = args.every(arg => {
+            if (Array.isArray(arg) === true) {
                 return !!arg.length;
-            }
-            // If an empty object is passed, arg is false
-            else if (_.isEmpty(arg) && _.isObject(arg)) {
+            } else if (arg && arg.constructor === Object && Object.keys(arg).length === 0) {
```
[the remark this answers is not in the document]
mattolson: I would also pull this and any other utility methods that are common throughout this PR into a library module.

helpers/any.js
```diff
@@ -24,22 +22,24 @@ const factory = () => {
         opts = args.pop();
         predicate = opts.hash;
 
-        if (!_.isEmpty(predicate)) {
+        if (Object.keys(predicate).length !== 0) {
```
mattolson: `> 0` is probably clearer here

helpers/any.js
```diff
@@ -24,22 +22,24 @@ const factory = () => {
...
             // With options hash, we check the contents of first argument
-            any = _.any(args[0], predicate);
+            any = args[0].some(arg => {
```
mattolson: `args[0]` could be either an object or an array, but we are assuming it's an array here. The lodash implementation supports both: https://github.com/lodash/lodash/blob/3.10.1/lodash.js#L7044

mattolson: Here's some documentation on this helper to help produce a correct implementation: https://developer.bigcommerce.com/stencil-docs/handlebars-syntax-and-helpers/handlebars-helpers-reference/new-subpage/loop-control-flow-helpers#handlebars_any

helpers/contains.js
```diff
@@ -12,8 +10,14 @@ const _ = require('lodash');
 const factory = () => {
     return function() {
         var args = Array.prototype.slice.call(arguments, 0, -1),
```
mattolson: We need to avoid slicing the arguments array.

helpers/deprecated/pick.js
```diff
@@ -1,13 +1,18 @@
...
     /**
      * @deprecate
      */
-    return function() {
-        return _.pick.apply(null, arguments);
+    return function(obj, ...args) {
+        const pickedElements = {};
+
+        args.forEach(e => {
+            if (obj[e]) {
```
mattolson: You should probably use `hasOwnProperty` here in case the value is falsy.

helpers/getImage.js
```diff
@@ -10,14 +8,19 @@ const factory = globals => {
         var size;
         var width;
         var height;
-
-        if (!_.isPlainObject(image) || !_.isString(image.data) || image.data.indexOf('{:size}') === -1) {
+        function isObject(obj) {
```
mattolson: Let's move these into a common library module

mattolson:
Two references that can help guide this refactor:
* https://github.com/lodash/lodash/blob/3.10.1/lodash.js
* https://developer.bigcommerce.com/stencil-docs/handlebars-syntax-and-helpers/handlebars-helpers-reference/handlebars-full

helpers/all.js
```diff
@@ -21,20 +19,15 @@ const factory = () => {
         opts = args.pop();
 
         // Check if all the arguments are valid / truthy
-        result = _.all(args, function (arg) {
-            if (_.isArray(arg)) {
+        result = args.every(arg => {
+            if (Array.isArray(arg) === true) {
```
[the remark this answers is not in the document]
bc-williamkwon: ok will change.

helpers/all.js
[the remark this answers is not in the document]
bc-williamkwon: ok will change.

helpers/any.js
bc-williamkwon: ok will change.

helpers/any.js
bc-williamkwon: ok will change.

helpers/getImage.js
bc-williamkwon: ok will do.

helpers/deprecated/pick.js
bc-williamkwon: ok will do.

helpers/contains.js
bc-williamkwon: ok.  will change it here too.

mattolson:
Please rebase after your `arguments` PR which changed a lot of things.

helpers/lib/common.js
```diff
@@ -0,0 +1,13 @@
+function isObject(val) {
+    return {}.toString.call(val) === '[object Object]';
+}
+function isOptions(val) {
```
mattolson: Prefer extra line between functions

helpers/lib/common.js
```diff
@@ -0,0 +1,13 @@
...
+    return isObject(val) && val.fn;
+}
+function isString(val) {
+    return typeof val === 'string';
+}
+function isEmptyObject(val) {
+    return isObject(val) && Object.keys(val).length === 0;
+}
+module.exports = {isObject, isOptions, isString, isEmptyObject};
```
mattolson: Prefer a multiline object here
```
module.exports = {
  isObject,
  isOptions,
  isString,
  isEmptyObject
};
```

helpers/lib/fonts.js
```diff
@@ -18,34 +16,29 @@ const fontProviders = {
...
-            var collection = [],
-                familyHash = {};
+            const collection = [];
+            const familyHash = {};
 
-            _.each(fonts, function fontsIterator(font) {
-                var split = font.split('_'),
-                    familyKey = split[1],  // Eg: Open+Sans
-                    weights = split[2];    // Eg: 400,700italic
+            fonts.forEach(font => {
```
mattolson: I think forEach is not going to give us a performance benefit. We should use a plain for loop: https://medium.com/tech-tajawal/loops-performances-in-node-js-9fbccf2d6aa6

Similarly for any other usages of forEach.

helpers/lib/fonts.js
```diff
@@ -104,38 +96,41 @@ module.exports = function(format, themeSettings, handlebars, options) {
...
 
-        collectedFonts[provider].push(value);
+        collectedFonts[provider].push(themeSettings[setting]);
     });
 
     // Parse font strings based on provider
-    const parsedFonts = _.mapValues(collectedFonts, function(value, key) {
-        return fontProviders[key].parser(value);
-    });
+    const parsedFonts = Object.keys(collectedFonts).reduce((parsedFonts, font) => {
```
mattolson: reduce, while nice and easy to read, is going to be slower than a for loop: https://hackernoon.com/3-javascript-performance-mistakes-you-should-stop-doing-ebf84b9de951

bc-williamkwon: fixed.

helpers/lib/fonts.js
bc-williamkwon: fixed.

helpers/lib/common.js
bc-williamkwon: fixed.

helpers/lib/common.js
bc-williamkwon: fixed.

helpers/lib/common.js
```diff
@@ -0,0 +1,33 @@
...
+
+function isString(val) {
+    return typeof val === 'string';
+}
+
+function isEmptyObject(val) {
+    return isObject(val) && Object.keys(val).length === 0;
+}
+
+function truthCheck(val) {
```
mattolson: Let's call this `isTruthy` (that's kind of a semi-standard term for this)

helpers/all.js
```diff
@@ -12,22 +10,16 @@ const factory = () => {
...
-                return !!arg.length;
-            }
-            // If an empty object is passed, arg is false
-            else if (_.isEmpty(arg) && _.isObject(arg)) {
-                return false;
+        for (let i = 0; i < args.length; i++) {
+            if (!common.truthCheck(args[i])) {
+                result = false;
+                break;
+            } else if (i === args.length - 1) {
```
mattolson: A more readable way to do this would be:
```
let result = true;
for (let i = 0; i < args.length; i++) {
  result = result && common.isTruthy(args[i]);
  if (!result) {
    break;
  }
}
```

helpers/any.js
```diff
@@ -15,27 +14,10 @@ const factory = () => {
         const opts = args.pop();
         const predicate = opts.hash;
 
-        if (!_.isEmpty(predicate)) {
+        if (Object.keys(predicate).length > 0) {
```
mattolson: Let's use `!isEmptyObject` here.

helpers/any.js
```diff
@@ -15,27 +14,10 @@ const factory = () => {
...
             // With options hash, we check the contents of first argument
             any = _.any(args[0], predicate);
-        } else {
```
mattolson: Not sure we want to remove this. I don't know that we've announced this deprecation anywhere.

helpers/lib/fonts.js
```diff
@@ -18,35 +16,33 @@ const fontProviders = {
...
+                   continue;
                 }
-
-                if (_.isUndefined(weights)) {
+                if (typeof weights === 'undefined') {
                     weights = '';
                 }
-
-                if (!_.isArray(familyHash[familyKey])) {
+                if (Array.isArray(familyHash[familyKey]) === false) {
```
mattolson: Should just do `!Array.isArray`

helpers/lib/fonts.js
```diff
@@ -18,35 +16,33 @@ const fontProviders = {
...
                     familyHash[familyKey] = [];
                 }
-
                 weights = weights.split(',');
-
                 familyHash[familyKey].push(weights);
-                familyHash[familyKey] = _.uniq(_.flatten(familyHash[familyKey]));
-            });
+                familyHash[familyKey] = [...new Set([...familyHash[familyKey]])];
```
mattolson: Let's just use `Array.from(set)`

helpers/lib/fonts.js
```diff
@@ -84,58 +86,76 @@ const fontProviders = {
...
  */
 module.exports = function(format, themeSettings, handlebars, options) {
-    
+
     const collectedFonts = {};
-    _.each(themeSettings, function(value, key) {
-        //check that -font is on end of string but not start of string
+    const themeSettingsKeys = Object.keys(themeSettings);
+    for (let i = 0; i < themeSettingsKeys.length; i++) {
+        const setting = themeSettingsKeys[i];
```
mattolson: I would put
```
const key = themeSettingsKeys[i];
const value = themeSettings[key];
```
at the top of the loop and that will minimize changes below and probably make things easier to read

helpers/limit.js
```diff
@@ -10,10 +8,10 @@ const _ = require('lodash');
  */
 const factory = () => {
     return function(data, limit) {
-        if (_.isString(data)) {
+        if (typeof data === 'string') {
```
mattolson: Let's use `common.isString`

spec/helpers/any.js
```diff
@@ -48,57 +48,3 @@ describe('any helper (with option hash)', function() {
     });
 });
 
-
-describe('any helper (with multiple arguments)', function() {
```
mattolson: Should probably not remove this

helpers/or.js
```diff
@@ -12,29 +11,22 @@ const factory = () => {
...
-        // Without options hash, we check all the arguments
-        const any = _.any(args, function (arg) {
-            if (_.isArray(arg)) {
-                return !!arg.length;
-            }
-            // If an empty object is passed, arg is false
-            else if (_.isEmpty(arg) && _.isObject(arg)) {
-                return false;
+        for (let i = 0; i < args.length; i++) {
+            if (common.truthCheck(args[i])) {
```
mattolson: Similar to the above, perhaps

```
let any = false;
for (let i = 0; i < args.length; i++) {
  any = any || common.isTruthy(args[i]);
  if (any) {
    break;
  }
}
```

helpers/pluck.js
```diff
@@ -1,10 +1,21 @@
...
-        return _.pluck(collection, path);
+        const arr = [];
+        if (Array.isArray(collection)) {
+            for (let i = 0; i < collection.length; i++) {
+                arr.push(collection[i][path]);
+            }
+        } else if (common.isObject(collection)) {
+            const keys = Object.keys(collection);
+            for (let i = 0; i < keys.length; i++) {
+                arr.push(collection[keys[i]][path]);
```
mattolson: Don't we need some kind of existence check here? Otherwise, we'll push `undefined` into the array.

helpers/resourceHints.js
```diff
@@ -16,28 +15,28 @@ const factory = globals => {
...
         // Add font providers
-        const fontProviders = _.keys(getFonts('providerLists', globals.getThemeSettings(), globals.handlebars));
-        _.each(fontProviders, function(provider) {
-            if (typeof fontResources[provider] !== 'undefined') {
-                hosts = hosts.concat(fontResources[provider]);
+        const providers = Object.keys(getFonts('providerLists', globals.getThemeSettings(), globals.handlebars));
+
+        for (let i = 0; i < providers.length; i++) {
+            if (typeof fontResources[providers[i]] !== 'undefined') {
+                hosts.push(...fontResources[providers[i]]);
```
mattolson: Why not `hosts.concat`?

index.js
```diff
@@ -66,9 +64,10 @@ class HandlebarsRenderer {
         };
 
         // Register helpers with Handlebars
-        helpers.forEach(spec => {
+        for (let i = 0; i < helpers.length; i++) {
+            const spec = helpers[i];
             this.handlebars.registerHelper(spec.name, spec.factory(this.helperContext));
-        });
+        }
```
mattolson: Extra whitespace

helpers/resourceHints.js
bc-williamkwon: Mainly to make hosts a const.

bc-williamkwon: It returns a new array too so I think it's doing more work.

helpers/lib/common.js
bc-williamkwon: done.

helpers/all.js
bc-williamkwon: done.

helpers/any.js
bc-williamkwon: done.

helpers/lib/fonts.js
bc-williamkwon: done.

helpers/lib/fonts.js
bc-williamkwon: done.

helpers/lib/fonts.js
bc-williamkwon: done.

helpers/limit.js
bc-williamkwon: done.

helpers/or.js
bc-williamkwon: done.

index.js
bc-williamkwon: done.

helpers/pluck.js
bc-williamkwon: Based on documentation.  I thought it would be ok to not do that.  But I have changed it.

helpers/any.js
bc-williamkwon: Unremoved.

spec/helpers/any.js
bc-williamkwon: unremoved.

helpers/lib/fonts.js
```diff
@@ -84,58 +86,77 @@ const fontProviders = {
  * @returns {Object.<string, Array>|string}
  */
 module.exports = function(format, themeSettings, handlebars, options) {
-    
+
     const collectedFonts = {};
-    _.each(themeSettings, function(value, key) {
-        //check that -font is on end of string but not start of string
+    const themeSettingsKeys = Object.keys(themeSettings);
```
mattolson: This portion of the change has become really complex, and I'm pretty sure something is wrong with this implementation. Perhaps we should strip this part out and save it for another PR?

helpers/pluck.js
```diff
@@ -1,10 +1,25 @@
 'use strict';
-
-var _ = require('lodash');
+const common = require('./lib/common.js');
 
 const factory = () => {
     return function(collection, path) {
-        return _.pluck(collection, path);
+        const arr = [];
```
mattolson: What is the current behavior of this function? Should it return an object when passed an object, or is it supposed to always return an array?

helpers/resourceHints.js
```diff
@@ -16,28 +15,28 @@ const factory = globals => {
...
             }
-        });
+        }
+
+        for (let j = 0; j < hosts.length; j++) {
```
mattolson: Perhaps instead of a second loop, we should just do this as part of the first loop?

index.js
```diff
@@ -166,22 +165,20 @@ class HandlebarsRenderer {
      * @param {Object} A set of templates to register with handlebars
      */
     addTemplates(templates) {
-        _.each(templates, (template, path) => {
-            // Don't do this work twice, first one wins.
-            if (typeof this.handlebars.partials[path] !== 'undefined') {
+        const keys = Object.keys(templates);
+
+        for (let i = 0; i < keys.length; i++) {
+            const template = keys[i];
```
mattolson: This should be named `path`

index.js
```diff
@@ -166,22 +165,20 @@ class HandlebarsRenderer {
      * @param {Object} A set of templates to register with handlebars
      */
     addTemplates(templates) {
-        _.each(templates, (template, path) => {
-            // Don't do this work twice, first one wins.
-            if (typeof this.handlebars.partials[path] !== 'undefined') {
+        const keys = Object.keys(templates);
```
mattolson: This should be named `paths`

index.js
```diff
@@ -166,22 +165,20 @@ class HandlebarsRenderer {
...
+        for (let i = 0; i < keys.length; i++) {
+            const template = keys[i];
+            if (typeof this.handlebars.partials[template] !== 'undefined') {
                 return;
             }
-
-            // Check if it is a precompiled template
             try {
-                template = this._tryRestoringPrecompiled(template);
+                templates[template] = this._tryRestoringPrecompiled(templates[template]);
```
mattolson: Let's not mutate the incoming parameter. `template` should be a local variable declared inside the loop.

index.js
```diff
@@ -221,14 +218,17 @@ class HandlebarsRenderer {
      */
     getPreProcessor() {
         return templates => {
+            const keys = Object.keys(templates);
```
mattolson: Let's name this `paths`

helpers/resourceHints.js
bc-williamkwon: We don't have individual access to each host in the first.

helpers/pluck.js
bc-williamkwon: .pluck is now map in lodash and returns a mapped array.  I used both _.pluck and my own against test cases and they pass.

bc-williamkwon: I think it would be better to keep lodash for collection helpers.

index.js
bc-williamkwon: fixed.

index.js
bc-williamkwon: fixed.

index.js
bc-williamkwon: done.

index.js
bc-williamkwon: fixed.

index.js
```diff
@@ -166,22 +165,21 @@ class HandlebarsRenderer {
      * @param {Object} A set of templates to register with handlebars
      */
     addTemplates(templates) {
-        _.each(templates, (template, path) => {
-            // Don't do this work twice, first one wins.
+        const paths = Object.keys(templates);
+
+        for (let i = 0; i < paths.length; i++) {
+            const path = paths[i];
+            let template;
```
mattolson: I would move this closer to its usage, i.e. right below the comment `// Check if it is a precompiled template`

helpers.js
```diff
@@ -3,21 +3,23 @@
...
 
 // Load helpers
-fs.readdirSync(Path.join(__dirname, 'helpers')).forEach(filename => {
-    if (!fs.lstatSync(Path.join(__dirname, 'helpers', filename)).isDirectory()) {
-        helpers = helpers.concat(require('./helpers/' + filename));
-    }
-});
+const notDeprecated = fs.readdirSync(Path.join(__dirname, 'helpers'));
 
+for (let i = 0; i < notDeprecated.length; i++) {
```
mattolson: Not important to refactor this file. This only gets loaded once, and the incremental gain of using a plain for loop is negligible compared to loading files from the file system. Let's revert the changes to this file.

helpers/all.js
```diff
@@ -12,22 +10,14 @@ const factory = () => {
     return function(...args) {
         // Take the last arg (which is a Handlebars options object) out of args array
         const opts = args.pop();
-
+        let result;
```
mattolson: Minor nitpick, but I would put this below the comment

helpers/lib/fonts.js
```diff
@@ -18,35 +18,33 @@ const fontProviders = {
...
+                if (!Array.isArray(familyHash[familyKey])) {
                     familyHash[familyKey] = [];
                 }
-
                 weights = weights.split(',');
-
                 familyHash[familyKey].push(weights);
-                familyHash[familyKey] = _.uniq(_.flatten(familyHash[familyKey]));
-            });
+                familyHash[familyKey] = Array.from(new Set([...familyHash[familyKey]]));
```
mattolson: This could probably be simplified

Instead of
```
familyHash[familyKey].push(weights);
familyHash[familyKey] = Array.from(new Set([...familyHash[familyKey]]));
```

how about
```
familyHash[familyKey].push(...weights);
```

and then in the final loop
```
for (let i = 0; i < familyHashKeys.length; i++) {
  const family = familyHashKeys[i];
  const weights = familyHash[family];
  const uniqueWeights = Array.from(new Set(weights));
  collection.push(`${family}:${uniqueWeights.join(',')}`);
}

```

index.js
```diff
@@ -166,22 +165,21 @@ class HandlebarsRenderer {
...
             if (typeof this.handlebars.partials[path] !== 'undefined') {
                 return;
             }
-
-            // Check if it is a precompiled template
```
mattolson: Please don't remove this comment

index.js
```diff
@@ -166,22 +165,21 @@ class HandlebarsRenderer {
...
             try {
-                template = this._tryRestoringPrecompiled(template);
+                template = this._tryRestoringPrecompiled(templates[path]);
             } catch(e) {
                 throw new FormatError(e.message);
             }
-
-            // Register it with handlebars
```
mattolson: Please don't remove this comment

index.js
bc-williamkwon: fixed

helpers.js
bc-williamkwon: fixed.

helpers/all.js
bc-williamkwon: fixed.

helpers/lib/fonts.js
bc-williamkwon: fixed.

index.js
bc-williamkwon: fixed.

index.js
bc-williamkwon: fixed.

helpers/all.js
bc-williamkwon: fixed.

helpers/lib/fonts.js
bc-williamkwon: fixed.

bc-williamkwon:
Removed lodash methods from any, contains, and pluck.

helpers/lib/common.js
```diff
@@ -0,0 +1,38 @@
...
+    if (Array.isArray(val)) {
+        return !!val.length;
+    } else if (isEmptyObject(val)) {
+        return false;
+    } else {
+        return !!val;
+    }
+}
+
+function sameValueZero(val1, val2) {
```
mattolson: I don't understand what this is trying to do. We probably need a comment here and probably a function name change.

bc-williamkwon: sameValueZero is a Lodash algorithm that also allows for NaN === NaN to return true.

bc-williamkwon: changed.

helpers/lib/fonts.js
```diff
@@ -18,35 +16,35 @@ const fontProviders = {
...
-            });
+            const familyHashKeys = Object.keys(familyHash);
 
-            _.each(familyHash, function fontHashIterator(weights, family) {
-                collection.push(family + ':' + weights.join(','));
-            });
+            for (let i = 0; i < familyHashKeys.length; i++) {
+                const family = familyHashKeys[i];
+                const weights = familyHash[family];
+                const uniqueWeights = Array.from(new Set(weights));
```
bc-zoharmuzafi: @bc-williamkwon you can just set in weights the unique values instead of defining a new var

helpers/lib/fonts.js
```diff
@@ -84,54 +88,76 @@ const fontProviders = {
...
-            return;
+            continue;
         }
 
         const splits = value.split('_');
         const provider = splits[0];
 
         if (typeof fontProviders[provider] === 'undefined') {
-            return;
+            continue;
```
bc-zoharmuzafi: Are we ok to change return to continue here?

helpers/resourceHints.js
```diff
@@ -16,28 +15,25 @@ const factory = globals => {
...
+        }
+        return new globals.handlebars.SafeString(hosts.join(''));
     };
 };
 
 module.exports = [{
     name: 'resourceHints',
     factory: factory,
-}];
+}];
```
bc-zoharmuzafi: new line

bc-zoharmuzafi:
@bc-williamkwon Overall the changes looks good. My concern is that this pr is changing very simple implementation and covers a lot of method. my suggestion due to the high risk, is to break this pr to really small prs, which cover one method each and to be sure each changed method backed up with sufficient unitest

helpers/lib/fonts.js
bc-williamkwon: yes, because return in a forEach is equivalent to continue in a for

PR #42, opened 2019-06-01
Add getImageSrcset helper

bookernath:
## What? Why?
Adding a helper to make it easy to get a `srcset` string for an image.

The API is meant to feel familiar to a user of `getImage`.

Intended HTML usage:

``

Output would look like:

`https://cdn00.bigcommerce.com/s-abc123/images/stencil/100w/l/my-huge-image.original.jpg 100w, https://cdn00.bigcommerce.com/s-abc123/images/stencil/200w/l/my-huge-image.original.jpg 200w, https://cdn00.bigcommerce.com/s-abc123/images/stencil/300w/l/my-huge-image.original.jpg 300w`

Or perhaps:
```
`<img src="{{getImage image "default"}}" srcset="{{getImageSrcset image 1x="1000x1000" 2x="2000x2000"}}" />`
```

For this result:

`https://cdn00.bigcommerce.com/s-abc123/images/stencil/1000x1000/l/my-huge-image.original.jpg 1x, https://cdn00.bigcommerce.com/s-abc123/images/stencil/2000x2000/l/my-huge-image.original.jpg 2x`

If you supply no arguments, the helper will use a default set of sizes.

Contains handling for default image (just like `getImage`).

## How was it tested?
Unit tests added. I've verified that it works against a [custom branch of Cornerstone](https://github.com/bigcommerce/cornerstone/pull/1507).

### DIFF
--- a/helpers/getImageSrcset.js
+++ b/helpers/getImageSrcset.js
@@ -1,0 +1,63 @@
+'use strict';
+const utils = require('handlebars-utils');
+const SafeString = require('handlebars').SafeString;
+const common = require('./lib/common');
+
+const factory = () => {
+    return function(image, defaultImageUrl) {
+        // Regex to test size string is of the form 123x123 or 100w
+        const sizeRegex = /(^\d+w$)|(^(\d+?)x(\d+?)$)/;
+        // Regex to test to that srcset descriptor is of the form 1x 1.5x 2x OR 123w
+        const descriptorRegex = /(^\d+w$)|(^([0-9](\.[0-9]+)?)x)$/;
+
+        const options = arguments[arguments.length - 1];
+
+        if (utils.isUndefined(defaultImageUrl)) {
+            defaultImageUrl = '';
+        }
+
+        if (!utils.isObject(image) || !utils.isString (image.data)
+            || !common.isValidURL(image.data) || image.data.indexOf('{:size}') === -1) {
+            // return empty string if not a valid image object
+            defaultImageUrl = defaultImageUrl ? defaultImageUrl : '';
+            return utils.isString(image) ? image : defaultImageUrl;
+        }
+
+        let srcsets = {};
+
+        if (options.hash['use_default_sizes']) {
+            srcsets = {
+                '2560w': '2560w',
+                '1920w': '1920w',
+                '1280w': '1280w',
+                '960w': '960w',
+                '640w': '640w',
+                '320w': '320w',
+                '160w': '160w',
+                '80w': '80w',
+            };
+        } else {
+            srcsets = options.hash;
+            if (!utils.isObject(srcsets) || Object.keys(srcsets).some(descriptor => {
+                return !(descriptorRegex.test(descriptor) && sizeRegex.test(srcsets[descriptor]));
+            })) {
+                // return empty string if not valid srcset object
+                return ''
+            }
+        }
+
+        // If there's only one argument, return a `src` only (also works for `srcset`)
+        if (Object.keys(srcsets).length === 1) {
+            return new SafeString((image.data.replace('{:size}', srcsets[Object.keys(srcsets)[0]])));
+        }
+
+        return new SafeString(Object.keys(srcsets).reverse().map(descriptor => {
+            return ([image.data.replace('{:size}', srcsets[descriptor]), descriptor].join(' '));
+        }).join(', '));
+    };
+};
+
+module.exports = [{
+    name: 'getImageSrcset',
+    factory: factory,
+}];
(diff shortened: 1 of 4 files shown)


helpers/getImageSrcset.js
```diff
@@ -0,0 +1,49 @@
...
+
+const factory = globals => {
+    return function(image, presetName, defaultImageUrl) {
+        // Regex to test size string is of the form 123x123
+        var sizeRegex = /^(\d+?)x(\d+?)$/;
+        // Regex to test to that srcset descriptor is of the form 1x 1.5x 2x OR 123w
+        var descriptorRegex = /(^\d+w$)|(^([0-9](\.[0-9]+)?)x)$/;
+        // Get theme settings from context
+        var settings = globals.getThemeSettings() || {};
+        var srcsets = settings[presetName] || [];
```
mattolson: To my knowledge, we don't support arrays or nested objects in the theme configuration right now -- just plain key/value pairs.

bookernath:
@mattolson ♻️ reworked this to use options on the helper itself instead of theme config.

helpers/getImageSrcset.js
```diff
@@ -0,0 +1,41 @@
...
+        const srcsets = options.hash;
+
+        if (!utils.isObject(srcsets) || Object.keys(srcsets).some(descriptor => {
+            return !(descriptorRegex.test(descriptor) && sizeRegex.test(srcsets[descriptor]));
+        })) {
+            // return empty string if not valid srcset object
+            return ''
+        }
+
+        return Object.keys(srcsets).reverse().map(descriptor => {
```
bookernath: I reverse the keys because the options hash returns them "backwards", which while harmless for a `srcset`, is a little unexpected by the person putting arguments on the handlebars helper.

helpers/getImageSrcset.js
bookernath: Moved to arguments per your suggestion.

helpers/getImageSrcset.js
```diff
@@ -0,0 +1,63 @@
...
+                return ''
+            }
+        }
+
+        // If there's only one argument, return a `src` only (also works for `srcset`)
+        if (Object.keys(srcsets).length === 1) {
+            return (image.data.replace('{:size}', srcsets[Object.keys(srcsets)[0]]));
+        }
+
+        return new SafeString(Object.keys(srcsets).reverse().map(descriptor => {
```
bookernath: I reverse the keys because the options hash returns them "backwards", which while harmless for a srcset, is a little unexpected by the person putting arguments on the handlebars helper.

helpers/getImageSrcset.js
```diff
@@ -0,0 +1,63 @@
...
+            return ([image.data.replace('{:size}', srcsets[descriptor]), descriptor].join(' '));
+        }).join(', '));
+    };
+};
+
+module.exports = [{
+    name: 'getImageSrcset',
```
junedkazi: Since this is internal helper to BC should we rename it to `getImageSrcSet` for consistency 🍹 ?

helpers/getImageSrcset.js
```diff
@@ -0,0 +1,63 @@
...
+                return !(descriptorRegex.test(descriptor) && sizeRegex.test(srcsets[descriptor]));
+            })) {
+                // return empty string if not valid srcset object
+                return ''
+            }
+        }
+
+        // If there's only one argument, return a `src` only (also works for `srcset`)
+        if (Object.keys(srcsets).length === 1) {
+            return (image.data.replace('{:size}', srcsets[Object.keys(srcsets)[0]]));
```
junedkazi: Do we need to run this through `SafeString` ?

helpers/getImageSrcset.js
```diff
@@ -0,0 +1,63 @@
...
+        const descriptorRegex = /(^\d+w$)|(^([0-9](\.[0-9]+)?)x)$/;
+
+        const options = arguments[arguments.length - 1];
+
+        if (utils.isUndefined(defaultImageUrl)) {
+            defaultImageUrl = '';
+        }
+
+        if (!utils.isObject(image) || !utils.isString (image.data)
+            || !common.isValidURL(image.data) || image.data.indexOf('{:size}') === -1) {
```
junedkazi: Well technically size should not be a requirement. It could be a theme asset where the theme developer already has  all the versions of the asset inside the theme. Something like this ```Elva dressed as a fairy```

bookernath: You don't need this helper then - this is for sizing the same images that are a candidate for `{{getImage}}`

helpers/getImageSrcset.js
bookernath: Yep, let me fix that

helpers/getImageSrcset.js
bookernath: Consistent with what? I'm kind of treating `srcset` as one word 🤷‍♂

junedkazi: I was thinking of it as 2 separate words 😆

PR #44, opened 2019-06-04
strf-6702 remove regex in font helpers

bc-williamkwon:
## What? Why?
STRF-6702 Remove the usage of the fontKeyFormat regex in the font helpers.  Regex is expensive and this is an alternative to that.

## How was it tested?
----
Unit tests pass.
Ran an analysis on regular expression analyzer and found that regex was equivalent to that there needs to be a word character before the occurrence of '-font' and '-font' needs to be end of string.  Word characters in ecmascript are [^a-zA-Z_0-9].  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Changelog
+## Unreleased
+- remove regex in fonts.js for better performance
 
 ## 4.0.6
 - Change default behavior of {{getFontsCollection}} to use font-display: swap for Google Fonts and allow configuration
--- a/helpers/lib/fonts.js
+++ b/helpers/lib/fonts.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const _ = require('lodash');
-const fontKeyFormat = new RegExp(/\w+(-\w*)*-font$/);
+
 const fontProviders = {
     'Google': {
         /**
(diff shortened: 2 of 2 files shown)


helpers/lib/fonts.js
```diff
@@ -87,8 +87,16 @@ module.exports = function(format, themeSettings, handlebars, options) {
     
     const collectedFonts = {};
     _.each(themeSettings, function(value, key) {
-        if (!fontKeyFormat.test(key)) {
+        //check that -font is on end of string but not start of string
```
mattolson: I think you are probably safe to use `key.endsWith('-font')`

bc-williamkwon: fixed.

mattolson: This seems unnecessarily complex. Could we just do the following?

```
const fontKeySuffix = '-font';
if (!key.endsWith(fontKeySuffix)) {
  return;
}
```

It's true that it's slightly different than the current regex, but I think it's safe. If the value doesn't follow the expected format, we will not treat it as a font either.

bc-williamkwon: changed.

mattolson:
Can you add something to the changelog?

bc-williamkwon:
added to changelog, should I merge?

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Changelog
+## 4.07
```
mattolson: I would call this "Unreleased", since we will probably wait for all these changes before cutting a new release

bc-williamkwon: fixed.

PR #47, opened 2019-06-05
STRF-6711 Remove cdn regex

bc-williamkwon:
## What? Why?
https://jira.bigcommerce.com/browse/STRF-6711 The regex usage in cdnify is fairly expensive. Figure out an alternative.  This pull request is for that alternative.

## How was it tested?
----
unit tests still pass.  Used regular expression analyzer to figure out what the regex statements were testing for.  Solution somewhat loosely translated can be more strict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 ## Unreleased
-- remove regex in fonts.js for better performance
+- Remove regex in cdnify.js to improve performance
+- Remove regex in fonts.js for better performance
 
 ## 4.0.6
 - Change default behavior of {{getFontsCollection}} to use font-display: swap for Google Fonts and allow configuration
--- a/helpers/lib/cdnify.js
+++ b/helpers/lib/cdnify.js
@@ -16,8 +16,6 @@ module.exports = globals => {
         const editSessionId = siteSettings.theme_session_id;
         const cdnSettings = themeSettings.cdn;
 
-        const protocolMatch = /(.*!?:)/;
-
         if (path instanceof globals.handlebars.SafeString) {
             path = path.string;
         }
@@ -26,28 +24,43 @@ module.exports = globals => {
             return '';
         }
 
-        if (/^(?:https?:)?\/\//.test(path)) {
+        if (path.startsWith('https://') || path.startsWith('http://') || path.startsWith('//')) {
             return path;
         }
-
+        // This block is to strip leading . and / from path
         if (['.', '/'].includes(path[0])) {
-            path = path.replace(/^[\.\/]+/, '');
+            for (let i = 0; i < path.length; i++) {
+                if (path[i] !== '.' && path[i] !== '/') {
+                    path = path.slice(i);
+                    break;
+                }
+            }
         }
 
-        if (protocolMatch.test(path)) {
-            var match = path.match(protocolMatch);
-            path = path.slice(match[0].length, path.length);
+        let colonIndex = path.indexOf(':');
 
+        if (colonIndex !== -1) {
+            while (colonIndex + 1 < path.length && path[colonIndex + 1] === ':') {
+                colonIndex++;
+            }
+            const protocol = path.slice(0, colonIndex + 1);
+            path = path.slice(colonIndex + 1);
+            //This block is to strip leading / from path
             if (path[0] === '/') {
-                path = path.slice(1, path.length);
+                for (let i = 1; i < path.length; i++) {
+                    if (path[i] !== '/') {
+                        path = path.slice(i);
+                        break;
+                    }
+                }
             }
 
-            if (match[0] === 'webdav:') {
+            if (protocol === 'webdav:') {
                 return [cdnUrl, 'content', path].join('/');
             }
 
             if (cdnSettings) {
-                var endpointKey = match[0].substr(0, match[0].length - 1);
+                const endpointKey = protocol.slice(0, protocol.length - 1);
                 if (cdnSettings.hasOwnProperty(endpointKey)) {
                     if (cdnUrl) {
                         return [cdnSettings[endpointKey], path].join('/');
@@ -64,7 +77,6 @@ module.exports = globals => {
             return path;
         }
 
-
         if (!versionId) {
             if (path[0] !== '/') {
                 path = '/' + path;
@@ -73,11 +85,11 @@ module.exports = globals => {
         }
 
         if (path[0] === '/') {
-            path = path.slice(1, path.length);
+            path = path.slice(1);
         }
 
-        if (path.match(/^assets\//)) {
-            path = path.substr(7, path.length);
+        if (path.startsWith('assets/')) {
+            path = path.slice(7);
         }
 
         if (editSessionId) {
--- a/spec/helpers/cdn.js
+++ b/spec/helpers/cdn.js
@@ -152,6 +152,10 @@ describe('cdn helper', function () {
                 input: '{{cdn "webdav:/img/image.jpg"}}',
                 output: 'https://cdn.bcapp/3dsf74g/content/img/image.jpg',
             },
+            {
+                input: '{{cdn "webdav://img/image.jpg"}}',
+                output: 'https://cdn.bcapp/3dsf74g/content/img/image.jpg',
+            },
         ], done);
     });
 


helpers/lib/cdnify.js
```diff
@@ -16,38 +16,45 @@ module.exports = globals => {
...
         if (path instanceof globals.handlebars.SafeString) {
             path = path.string;
         }
 
         if (!path) {
             return '';
         }
-
-        if (/^(?:https?:)?\/\//.test(path)) {
+
```
mattolson: Trailing whitespace

helpers/lib/cdnify.js
```diff
@@ -16,38 +16,45 @@ module.exports = globals => {
...
             path = path.string;
         }
 
         if (!path) {
             return '';
         }
-
-        if (/^(?:https?:)?\/\//.test(path)) {
+        
+        if (path.startsWith('//') || path.startsWith('http')) {
```
mattolson: Let's reverse the order here since it'll be more common for the url to start with `http` than `//` at this point

helpers/lib/cdnify.js
```diff
@@ -16,38 +16,45 @@ module.exports = globals => {
...
             return path;
         }
 
         if (['.', '/'].includes(path[0])) {
-            path = path.replace(/^[\.\/]+/, '');
+            for (var i = 0; i < path.length; i++) {
```
mattolson: Prefer `let` over `var`

helpers/lib/cdnify.js
```diff
@@ -16,38 +16,45 @@ module.exports = globals => {
...
+                if (path[i] !== '.' && path[i] !== '/') {
```
mattolson: This doesn't match current behavior. This block is meant to remove leading dots and slashes from the beginning of the string.

helpers/lib/cdnify.js
bc-williamkwon: I chose var here since let is block scoped and reinitialized at every iteration of the for loop so it will run slower.

helpers/lib/cdnify.js
bc-williamkwon: That is what the for loop is doing.

helpers/lib/cdnify.js
bc-williamkwon: fixed.

helpers/lib/cdnify.js
bc-williamkwon: fixed.

helpers/lib/cdnify.js
bc-williamkwon: nvm.  Looked it up some more and I guess what I read was outdated.

helpers/lib/cdnify.js
mattolson: Ok, I see now. Maybe adding some comments would help.

bc-williamkwon: fixed.

helpers/lib/cdnify.js
```diff
@@ -26,28 +24,38 @@ module.exports = globals => {
             return '';
         }
 
-        if (/^(?:https?:)?\/\//.test(path)) {
+        if (path.startsWith('http') || path.startsWith('//')) {
```
mattolson: I think to be safe with the refactor, we should be more explicit here:

```
if (path.startsWith('https://') || path.startsWith('http://') || path.startsWith('//'))
```

helpers/lib/cdnify.js
```diff
@@ -26,28 +24,38 @@ module.exports = globals => {
...
             return path;
         }
-
+        //this block is to strip leading . and / of path
```
mattolson: Minor formatting and word change request: `// This block is to strip leading . and / from path`

helpers/lib/cdnify.js
```diff
@@ -76,8 +83,8 @@ module.exports = globals => {
             path = path.slice(1, path.length);
         }
 
-        if (path.match(/^assets\//)) {
-            path = path.substr(7, path.length);
+        if (path.startsWith('assets/')) {
+            path = path.slice(7, path.length);
```
mattolson: We should probably just omit this second parameter for clarity, and for any similar usages in this file.

helpers/lib/cdnify.js
```diff
@@ -26,28 +24,38 @@ module.exports = globals => {
...
-        if (protocolMatch.test(path)) {
-            var match = path.match(protocolMatch);
-            path = path.slice(match[0].length, path.length);
+        let colonIndex = path.indexOf(':');
+
+        if (colonIndex !== -1) {
+            while (path[colonIndex + 1] === ':') {
+                colonIndex++;
+            }
+            const match = path.slice(0, colonIndex + 1);
```
mattolson: I would name this `protocol` for clarity. We are no longer using a regex which returns a match array, so "match" isn't very clear.

helpers/lib/cdnify.js
bc-williamkwon: done.

helpers/lib/cdnify.js
bc-williamkwon: done.

helpers/lib/cdnify.js
bc-williamkwon: done.

helpers/lib/cdnify.js
bc-williamkwon: done.

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Changelog
+## 4.0.7
```
mattolson: I would call this "Unreleased"

helpers/lib/cdnify.js
```diff
@@ -26,28 +24,38 @@ module.exports = globals => {
...
+            }
         }
 
-        if (protocolMatch.test(path)) {
-            var match = path.match(protocolMatch);
-            path = path.slice(match[0].length, path.length);
+        let colonIndex = path.indexOf(':');
+
+        if (colonIndex !== -1) {
+            while (path[colonIndex + 1] === ':') {
```
mattolson: You probably need an additional guard here to avoid out of bounds errors: `while (colonIndex + 1 < path.length && path[colonIndex + 1] === ':')`

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
...
+- Remove regex in cdnify.js to improve performance.
```
mattolson: I would call this "Unreleased"

helpers/lib/cdnify.js
bc-williamkwon: with Javascript there wouldn't be an out of bounds error just return undefined.  Which would break the loop.  If you then slice the string it would be an empty string.

bc-williamkwon: Just realized we wouldn't actually slice the string if that were the case.

bc-williamkwon: Err.  Just realized I thought we were thinking of some other piece of code instead of this one.

CHANGELOG.md
bc-williamkwon: fixed

helpers/lib/cdnify.js
```diff
@@ -26,28 +24,38 @@ module.exports = globals => {
...
-            var match = path.match(protocolMatch);
-            path = path.slice(match[0].length, path.length);
+        let colonIndex = path.indexOf(':');
+
+        if (colonIndex !== -1) {
+            while (colonIndex + 1 < path.length && path[colonIndex + 1] === ':') {
+                colonIndex++;
+            }
+            const protocol = path.slice(0, colonIndex + 1);
+            path = path.slice(colonIndex + 1);
```
mattolson: Tracing through this code with path = "webdav://foobar", it seems that at this point we will be left with "//foobar" and then in the next step, "/foobar", and then we will have too many "/" in the response.

bc-williamkwon: It seems like this was a bug in the old way too.  Will fix.

bc-williamkwon: fixed.

bc-williamkwon:
Fixed conflict and added a test for webdav://

bc-williamkwon:
Test coverage looks sufficient.

mattolson:
Please squash commits

bc-williamkwon:
Sorry, things are squashed now.  For some reason I thought that the merge conflict fix commit would be ignored when merged.

PR #48, opened 2019-06-09
STRF-6707 Don't use arguments in helpers

bc-williamkwon:
## What? Why?
STRF 6707 Don't use arguments in helpers.  Move away from arguments pattern so that performance may be improved.

## How was it tested?

----
Added unit test for for helper fix and current unit tests don't break.

### DIFF
--- a/helpers/all.js
+++ b/helpers/all.js
@@ -9,19 +9,12 @@ const _ = require('lodash');
  * {{#all items theme_settings.optionA theme_settings.optionB}} ... {{/all}}
  */
 const factory = () => {
-    return function() {
-        var args = [], opts, result;
-
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
-
-        // Take the last argument (content) out of testing array
-        opts = args.pop();
+    return function(...args) {
+        // Take the last arg (which is a Handlebars options object) out of args array
+        const opts = args.pop();
 
         // Check if all the arguments are valid / truthy
-        result = _.all(args, function (arg) {
+        const result = _.all(args, function (arg) {
             if (_.isArray(arg)) {
                 return !!arg.length;
             }
--- a/helpers/any.js
+++ b/helpers/any.js
@@ -9,20 +9,11 @@ const _ = require('lodash');
  * {{#any items selected=true}} ... {{/any}}
  */
 const factory = () => {
-    return function() {
-        var args = [],
-            opts,
-            predicate,
-            any;
-
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
-
-        // Take the last argument (content) out of testing array
-        opts = args.pop();
-        predicate = opts.hash;
+    return function(...args) {
+        let any;
+        // Take the last arg (which is a Handlebars options object) out of args array
+        const opts = args.pop();
+        const predicate = opts.hash;
 
         if (!_.isEmpty(predicate)) {
             // With options hash, we check the contents of first argument
--- a/helpers/block.js
+++ b/helpers/block.js
@@ -1,10 +1,9 @@
 'use strict';
 
 const factory = globals => {
-    return function(name) {
-        const options = arguments[arguments.length - 1];
-
+    return function(name, options) {
         /* Look for partial by name. */
+        options = options ? options : name;
         const partial = globals.handlebars.partials[name] || options.fn;
         return partial(this, { data: options.hash });
     };
--- a/helpers/compare.js
+++ b/helpers/compare.js
@@ -1,19 +1,14 @@
 'use strict';
 
 const factory = () => {
-    return function(lvalue, rvalue) {
-        const options = arguments[arguments.length - 1];
-        var operator;
-        var operators;
-        var result;
-
-        if (arguments.length < 3) {
-            throw new Error("Handlerbars Helper 'compare' needs 2 parameters");
+    return function(lvalue, rvalue, options) {
+        if (typeof options === 'undefined') {
+            throw new Error("Handlerbars Helper 'compare' needs 2 parameters and a block.");
         }
 
-        operator = options.hash.operator || "==";
+        const operator = options.hash.operator || "==";
 
-        operators = {
+        const operators = {
             '==': function (l, r) { return l == r; },
             '===': function (l, r) { return l === r; },
             '!=': function (l, r) { return l != r; },
@@ -32,7 +27,7 @@ const factory = () => {
             throw new Error("Handlerbars Helper 'compare' doesn't know the operator " + operator);
         }
 
-        result = operators[operator](lvalue, rvalue);
+        const result = operators[operator](lvalue, rvalue);
 
         if (result) {
             return options.fn(this);
--- a/helpers/contains.js
+++ b/helpers/contains.js
@@ -10,11 +10,8 @@ const _ = require('lodash');
  * {{#contains font_path "Roboto"}} ... {{/contains}}
  */
 const factory = () => {
-    return function() {
-        var args = Array.prototype.slice.call(arguments, 0, -1),
-            options = _.last(arguments),
-            contained = _.contains.apply(_, args);
-
+    return function(container, value, options) {
+        const contained = _.contains(container, value);
         // Yield block if true
         if (contained) {
             return options.fn(this);
--- a/helpers/for.js
+++ b/helpers/for.js
@@ -3,21 +3,22 @@
 const _ = require('lodash');
 
 const factory = () => {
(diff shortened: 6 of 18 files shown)


helpers/any.js
```diff
@@ -9,20 +9,11 @@ const _ = require('lodash');
...
-            any;
-
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
-
+    return function(...args) {
+        let any;
         // Take the last argument (content) out of testing array
```
mattolson: This comment doesn't seem to make sense anymore...

helpers/for.js
```diff
@@ -3,29 +3,30 @@
...
+            options = context;
             if (_.isObject(to)) {
                 context = to;
                 to = from;
                 from = 1;
             }
         }
 
-        if (to <= from) {
+        if (to < from) {
```
mattolson: Why was this changed? Seems like this will alter current looping logic.

helpers/or.js
```diff
@@ -9,21 +9,12 @@ const _ = require('lodash');
...
-            opts,
-            any;
-
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
-
+    return function(...args) {
         // Take the last argument (content) out of testing array
```
mattolson: This comment seems incorrect now

helpers/for.js
bc-williamkwon: It allows for one iteration.  Previously we could only do two iterations at the minimum.  There was an issue written about it too.

helpers/or.js
bc-williamkwon: I thought it made sense as arguments (non-keyword) are passed to functions.

helpers/for.js
mattolson: Let's pull this out as a separate change so we can evaluate it separately.

helpers/or.js
mattolson: Maybe if it said "options" instead of "content", it would make more sense.

bc-williamkwon: fixed

bc-williamkwon: Nvm rebasing issues . . .

bc-williamkwon: Git is behaving weird but I believe things are fixed now.

bc-williamkwon: whitespace in test case of for.js is fixed in the other pr.

bc-williamkwon: fixed here now too actually.

bc-williamkwon: Actually realized that I missed a file.

bc-williamkwon: Three files have been changed to be more explicit about what is being pulled out of the args array.

bc-williamkwon:
The rest parameter pattern (...args) is more performant when processing things as an array but otherwise it's better to use named parameters.  I changed contains and stylesheet based on documentation ```contains Helper
The contains helper is custom to Stencil. It checks whether the second parameter is included in the first parameter (typically a collection).```
```{{stylesheet}}
The stylesheet helper is custom to Stencil. It renders a link tag to insert a stylesheet into your theme. (This is required if you want Theme Editor to rewrite the stylesheet file when a merchant customizes their theme.) This helper returns an HTML string.

Parameters
path: String containing the path to the theme’s CSS stylesheet file.
Other parameters are optional, appended in the form: key="value".
Example
{{{stylesheet "assets/css/style.css" class="myStylesheet"}}}```

Also partial.js is called in the form of Handlebars.registerPartial('myPartial', '{{name}}') so I think this should be safe as well.

File: helpers/all.js [hunk 4]
```diff
@@ -9,19 +9,12 @@ const _ = require('lodash');
  * {{#all items theme_settings.optionA theme_settings.optionB}} ... {{/all}}
  */
 const factory = () => {
-    return function() {
```
mattolson: I'm going to start leaving comments for each helper individually to help me keep track of what portions are approved or still under review.

Changes to this helper: approved.

helpers/any.js
```diff
@@ -9,20 +9,11 @@ const _ = require('lodash');
  * {{#any items selected=true}} ... {{/any}}
  */
 const factory = () => {
-    return function() {
```
mattolson: Changes to this helper: approved

helpers/compare.js
```diff
@@ -1,19 +1,14 @@
 'use strict';
 
 const factory = () => {
-    return function(lvalue, rvalue) {
```
mattolson: Changes to this helper: approved

helpers/block.js
```diff
@@ -1,10 +1,9 @@
 'use strict';
 
 const factory = globals => {
-    return function(name) {
```
mattolson: Changes to this helper: approved

helpers/contains.js
```diff
@@ -10,11 +10,8 @@ const _ = require('lodash');
  * {{#contains font_path "Roboto"}} ... {{/contains}}
  */
 const factory = () => {
-    return function() {
```
mattolson: Changes to this helper: approved

helpers/for.js
```diff
@@ -3,21 +3,22 @@
 const _ = require('lodash');
 
 const factory = () => {
-    return function(from, to, context) {
```
mattolson: Changes to this helper: approved

helpers/join.js
```diff
@@ -1,9 +1,8 @@
 'use strict';
 
 const factory = () => {
-    return function(array, separator) {
```
mattolson: Changes to this helper: approved

helpers/lang.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 const factory = globals => {
-    return function(translationKey) {
```
mattolson: Changes to this helper: approved

helpers/unless.js
```diff
@@ -1,15 +1,15 @@
 'use strict';
 
 const factory = globals => {
-    return function() {
```
mattolson: Changes to this helper: approved

helpers/stylesheet.js
```diff
@@ -4,13 +4,11 @@ const _ = require('lodash');
 const buildCDNHelper = require('./lib/cdnify');
 
 const factory = globals => {
-    return function(assetPath) {
+    return function(assetPath, options) {
```
mattolson: This doesn't allow for html attributes that don't have values (i.e. not passed in via options.hash)

helpers/replace.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 const factory = () => {
-    return function(needle, haystack) {
```
mattolson: Changes to this helper: approved

helpers/partial.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 const factory = globals => {
-    return function(name) {
```
mattolson: Changes to this helper: approved

helpers/or.js
```diff
@@ -9,21 +9,12 @@ const _ = require('lodash');
  * {{#or 1 0 0 0 0 0}} ... {{/or}}
  */
 const factory = () => {
-    return function() {
```
mattolson: Changes to this helper: approved

helpers/if.js
```diff
@@ -3,8 +3,7 @@
 const _ = require('lodash');
 
 const factory = globals => {
-    return function(lvalue, operator, rvalue) {
```
mattolson: Changes to this helper: approved

helpers/getFontsCollection.js
```diff
@@ -3,9 +3,8 @@
 const getFonts = require('./lib/fonts');
 
 const factory = globals => {
-    return function() {
```
mattolson: Changes to this helper: approved

spec/helpers/stylesheet.js
```diff
@@ -25,7 +25,7 @@ describe('stylesheet helper', () => {
     it('should render a link tag and all extra attributes with no cdn url', done => {
         runTestCases([
             {
-                input: '{{{stylesheet "assets/css/style.css" blah rel="something" class="myStyle"}}}',
```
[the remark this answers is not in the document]
bc-williamkwon: ok, thanks for your insight.  I will fix it.

bc-williamkwon:
rebased and fixed merge conflicts.

helpers/stylesheet.js
```diff
@@ -4,13 +4,11 @@ const _ = require('lodash');
 const buildCDNHelper = require('./lib/cdnify');
 
 const factory = globals => {
-    return function(assetPath) {
```
mattolson: Changes to this helper: approved

bc-williamkwon:
Thanks for doing the code review.

PR #49, opened 2019-06-10
STRF-6836 For helper should iterate once

### LINKED ISSUE #28: 'for' helper should run if 'from' and 'to' are equal
I've reported this before, but just came across it again, so here it goes:

https://github.com/bigcommerce/paper/issues/90

It is currently impossible to run the `for` helper one time. It's coded not to run if `from <= to`. However, the `for` loop in the helper runs from `i = from` to `i <= to`. Either the for loop needs to change, or the check on line 28 should be changed to `if (to < from) {`:

https://github.com/bigcommerce/paper-handlebars/blob/master/helpers/for.js#L28

In other words, if `from = 1` and `to = 2`, the helper runs twice. But if `from = 1` and `to = 1`, the helper doesn't run at all. It needs to be updated so it can run once if necessary.

bc-williamkwon:
## What? Why?
STRF-6836 For helper should iterate once.  Previously for helper would iterate twice at the minimum and now it is able to iterate once at minimum.

Fixes #28 

## How was it tested?

----
Added unit tests and tests pass.

### DIFF
--- a/helpers/for.js
+++ b/helpers/for.js
@@ -26,7 +26,7 @@ const factory = () => {
             }
         }
 
-        if (to <= from) {
+        if (to < from) {
             return;
         }
 
--- a/spec/helpers/for.js
+++ b/spec/helpers/for.js
@@ -10,6 +10,27 @@ describe('for helper', function() {
     // Build a test runner
     const runTestCases = testRunner({context});
 
+    it('should iterate once.', function(done) {
+        runTestCases([
+            {
+                input: '{{#for 1 this}}{{$index}}:{{name}} {{/for}}',
+                output: '1:Joe ',
+            },
+            {
+                input: '{{#for 1 1 this}}{{$index}}:{{name}} {{/for}}',
+                output: '1:Joe ',
+            },
+            {
+                input: '{{#for 0 0 this}}{{$index}}:{{name}} {{/for}}',
+                output: '0:Joe ',
+            },
+            {
+                input: '{{#for 1000 1000 this}}{{$index}}:{{name}} {{/for}}',
+                output: '1000:Joe ',
+            },
+        ], done);
+    });
+
     it('should iterate 10 times', function(done) {
         runTestCases([
             {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Remove regex in cdnify.js to improve performance
 - Remove regex in fonts.js for better performance
 - Refactor functions away from arguments pattern for better performance
+- Fix for helper to allow for one iteration
 
 ## 4.0.6
 - Change default behavior of {{getFontsCollection}} to use font-display: swap for Google Fonts and allow configuration


bc-williamkwon:
I added a bit of code in the case that we decide to support this alternate behavior by checking the options.hash for zeroDiffStart.

helpers/for.js
```diff
@@ -3,30 +3,37 @@
...
             if (_.isObject(to)) {
                 context = to;
                 to = from;
                 from = 1;
             }
         }
 
-        if (to <= from) {
-            return;
+        if (options.hash['zeroDiffStart'] !== true) {
```
mattolson: Let's change the option name to `includeLast` which I think will make more sense to people.

spec/helpers/for.js
```diff
@@ -10,6 +10,48 @@ describe('for helper', function() {
     // Build a test runner
     const runTestCases = testRunner({context});
 
+    it('should iterate once if zeroDiffStart is true.', function(done) {
```
mattolson: Would be more descriptive to say `includes the last element if includeLast is true`

spec/helpers/for.js
```diff
@@ -10,6 +10,48 @@ describe('for helper', function() {
...
+                output: '0:Joe ',
+            },
+            {
+                input: '{{#for 1000 1000 this zeroDiffStart=true}}{{$index}}:{{name}} {{/for}}',
+                output: '1000:Joe ',
+            },
+        ], done);
+    });
+
+    it('should not iterate once if zeroDiffStart is not set to true.', function(done) {
```
mattolson: Would be more descriptive to say `does not include the last element if includeLast is not true`

spec/helpers/for.js
```diff
@@ -10,6 +10,48 @@ describe('for helper', function() {
...
+                output: '1:Joe ',
+            },
+            {
+                input: '{{#for 0 0 this zeroDiffStart=true}}{{$index}}:{{name}} {{/for}}',
+                output: '0:Joe ',
+            },
+            {
+                input: '{{#for 1000 1000 this zeroDiffStart=true}}{{$index}}:{{name}} {{/for}}',
+                output: '1000:Joe ',
+            },
```
mattolson: I'd add one more case of `for 1 2` just to make this even more clear that this applies to more than just a single iteration.

mattolson:
Hmm, I think I was mistaken with my previous comments.

Reading the code more closely, there is a mismatch in the logic on line 29 and line 40. Unfortunately, the developer documentation doesn’t specify if it’s meant to be inclusive or exclusive of the last index. Since in the majority of cases, it is inclusive, we should just change the guard clause on line 29 and get rid of the new `zeroDiffStart` option.

[The only relevant test for this](https://github.com/bigcommerce/paper-handlebars/blob/master/spec/helpers/for.js#L96) doesn't indicate the desired behavior when `from == to` either.

This is a possible breaking change for anyone running loops with a single iteration, but I would imagine that this is the expected behavior.

PR #50, opened 2019-06-17
STRF-6701 Renderer: reduce usage of Lodash

bc-williamkwon:
## What? Why?
STRF-6701 Renderer: Reduce usage of Lodash.  Lodash utility methods can be slow especially things like _.each and this is a pr to improve performance of Storefront Renderer by moving to native JavaScript methods and for loops.

## How was it tested?

----
Unit tests still pass.

### DIFF
--- a/helpers/getImage.js
+++ b/helpers/getImage.js
@@ -1,23 +1,20 @@
 'use strict';
-
-var _ = require('lodash');
+const common = require('./lib/common.js');
 
 const factory = globals => {
     return function(image, presetName, defaultImageUrl) {
-        var sizeRegex = /^(\d+?)x(\d+?)$/g;
-        var settings = globals.getThemeSettings() || {};
-        var presets = settings._images;
-        var size;
-        var width;
-        var height;
+        const sizeRegex = /^(\d+?)x(\d+?)$/g;
+        const settings = globals.getThemeSettings() || {};
+        const presets = settings._images;
+        let size, width, height;
 
-        if (!_.isPlainObject(image) || !_.isString(image.data) || image.data.indexOf('{:size}') === -1) {
+        if (!common.isObject(image) || !common.isString(image.data) || image.data.indexOf('{:size}') === -1) {
             // return empty string if not a valid image object
             defaultImageUrl = defaultImageUrl ? defaultImageUrl : '';
-            return _.isString(image) ? image : defaultImageUrl;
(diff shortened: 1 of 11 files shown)


helpers/getImage.js
```diff
@@ -1,23 +1,22 @@
...
-        var presets = settings._images;
-        var size;
-        var width;
-        var height;
-
-        if (!_.isPlainObject(image) || !_.isString(image.data) || image.data.indexOf('{:size}') === -1) {
+        const sizeRegex = /^(\d+?)x(\d+?)$/g;
+        const settings = globals.getThemeSettings() || {};
+        const presets = settings._images;
+        let size;
```
mattolson: Prefer one line for this: `let size, width, height;`

helpers/lib/common.js
```diff
@@ -0,0 +1,38 @@
...
+        return !!val.length;
+    } else if (isEmptyObject(val)) {
+        return false;
+    } else {
+        return !!val;
+    }
+}
+//function to check if two values are equal also does a check for NaN === NaN
+function equals(val1, val2) {
+    return val1 === val2 || (val1 !== val1 && val2 !== val2);
```
mattolson: Can we use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is for this?

bc-williamkwon: Learn something new everyday.  I believe we can.

helpers/getImage.js
bc-williamkwon: ok, changed.

helpers/lib/common.js
bc-williamkwon: fixed.

helpers/lib/common.js
```diff
@@ -0,0 +1,38 @@
...
+    if (Array.isArray(val)) {
+        return !!val.length;
+    } else if (isEmptyObject(val)) {
+        return false;
+    } else {
+        return !!val;
+    }
+}
+//function to check if two values are equal also does a check for NaN === NaN
+function equals(val1, val2) {
```
mattolson: I think now that we have an equivalence for this, we really don't need this method anymore, let's just use `Object.is` whenever we need that behavior. I don't see a need for it in the current PR.

bc-williamkwon: I have removed the function.

PR #51, opened 2019-06-17
STRF-6701 Renderer: reduce usage of Lodash

bc-williamkwon:
## What? Why?
STRF-6701 Renderer: Reduce usage of Lodash. Lodash utility methods can be slow especially things like _.each and this is a pr to improve performance of Storefront Renderer by moving to native JavaScript methods and for loops.

## How was it tested?

----
The unit tests still pass.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,4 @@
 'use strict';
-
-const _ = require('lodash');
 const HandlebarsV3 = require('handlebars');
 const HandlebarsV4 = require('@bigcommerce/handlebars-v4');
 const helpers = require('./helpers');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -166,22 +165,23 @@ class HandlebarsRenderer {
...
+            const path = paths[i];
+
             if (typeof this.handlebars.partials[path] !== 'undefined') {
                 return;
             }
-
             // Check if it is a precompiled template
             try {
-                template = this._tryRestoringPrecompiled(template);
+                const template = this._tryRestoringPrecompiled(templates[path]);
```
mattolson: You should move the comment down if you want to handle it this way.

```
try {
  // Check if it is a precompiled template
  const template = this._tryRestoringPrecompiled(templates[path]);

  // Register it with handlebars
  this.handlebars.registerPartial(path, template);
} catch(e) {
}
```

mattolson: Or if you prefer

```
try {
  // Register it with handlebars after checking that it is a precompiled template
  const template = this._tryRestoringPrecompiled(templates[path]);
  this.handlebars.registerPartial(path, template);
} catch(e) {
}
```

bc-williamkwon: I went the first way.  I thought it kind of looked weird before and this makes sense.

PR #52, opened 2019-06-17
STRF-6701 Renderer: reduce Lodash usage

bc-williamkwon:
## What? Why?
STRF-6701 Renderer: Reduce usage of Lodash. Lodash utility methods can be slow especially things like _.each and this is a pr to improve performance of Storefront Renderer by moving to native JavaScript methods and for loops.

## How was it tested?

----
Unit tests still pass also added unit tests for object iteration.

### DIFF
--- a/helpers/any.js
+++ b/helpers/any.js
@@ -1,7 +1,6 @@
 'use strict';
 
-const _ = require('lodash');
-
+const common = require('./lib/common.js');
 /**
  * Yield block if any object within a collection matches supplied predicate
  *
@@ -15,25 +14,47 @@ const factory = () => {
         const opts = args.pop();
         const predicate = opts.hash;
 
-        if (!_.isEmpty(predicate)) {
+        if (!common.isEmptyObject(predicate)) {
             // With options hash, we check the contents of first argument
-            any = _.any(args[0], predicate);
+            const predicateKeys = Object.keys(predicate);
+            const arg = args[0];
+
+            function matchesPredicate(element, predicate) {
+                for (let i = 0; i < predicateKeys.length; i++) {
+                    if (!Object.is(element[predicateKeys[i]], predicate[predicateKeys[i]])) {
+                        return false;
+                    }
+                }
+                return true;
+            }
+
+            if (Array.isArray(arg)) {
+                for (let i = 0; i < arg.length; i++) {
+                    const element = arg[i];
+                    any = matchesPredicate(element, predicate);
+                    if (any) {
+                        break;
+                    } 
+                }
+            } else if (common.isObject(arg)) {
+                const argIndices = Object.keys(arg);
+                for (let i = 0; i < argIndices.length; i++) {
+                    const element = arg[argIndices[i]];
+                    any = matchesPredicate(element, predicate);
+                    if (any) {
+                        break;
+                    }
+                }
+            }
         } else {
             // DEPRECATED: Moved to #or helper
             // Without options hash, we check all the arguments
-            any = _.any(args, function (arg) {
-                if (_.isArray(arg)) {
-                    return !!arg.length;
+            for (let i = 0; i < args.length; i++) {
+                any = common.isTruthy(args[i]);
+                if (any) {
+                    break;
                 }
-                // If an empty object is passed, arg is false
-                else if (_.isEmpty(arg) && _.isObject(arg)) {
-                    return false;
-                }
-                // Everything else
-                else {
-                    return !!arg;
-                }
-            });
+            }
         }
 
         if (any) {
--- a/spec/helpers/any.js
+++ b/spec/helpers/any.js
@@ -7,13 +7,16 @@ const Lab = require('lab'),
 describe('any helper (with option hash)', function() {
     const context = {
         big: 'big',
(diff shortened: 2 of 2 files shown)


helpers/any.js
```diff
@@ -1,7 +1,7 @@
-'use strict';
 
-const _ = require('lodash');
+'use strict';
```
mattolson: This should be the first line in the file

helpers/any.js
```diff
@@ -15,25 +15,37 @@ const factory = () => {
         const opts = args.pop();
         const predicate = opts.hash;
 
-        if (!_.isEmpty(predicate)) {
+        if (!common.isEmptyObject(predicate)) {
             // With options hash, we check the contents of first argument
-            any = _.any(args[0], predicate);
+            const predicateKey = Object.keys(predicate)[0];
```
mattolson: We only check the first key from the options hash? The lodash docs seem to imply otherwise: https://lodash.com/docs/2.4.2#some

bc-williamkwon: So this is another thing that has a disconnect with our docs (the other being pluck) where the example use is selected=true and we don't use more than one predicate in the cornerstone theme.

helpers/any.js
bc-williamkwon: whoops.  Fixed.

helpers/any.js
```diff
@@ -15,25 +14,48 @@ const factory = () => {
...
-        if (!_.isEmpty(predicate)) {
+        if (!common.isEmptyObject(predicate)) {
             // With options hash, we check the contents of first argument
-            any = _.any(args[0], predicate);
+            const predicateKeys = Object.keys(predicate);
+            const arg = args[0];
+            if (Array.isArray(arg)) {
+                for (let i = 0; i < arg.length; i++) {
+                    const element = arg[i];
+                    for (let j = 0; j < predicateKeys.length; j++) {
```
mattolson: This inner portion is common to both cases so let's pull it out into a common function.

bc-williamkwon: fixed.

helpers/any.js
```diff
@@ -15,25 +14,37 @@ const factory = () => {
...
-        if (!_.isEmpty(predicate)) {
+        if (!common.isEmptyObject(predicate)) {
             // With options hash, we check the contents of first argument
-            any = _.any(args[0], predicate);
+            const predicateKeys = Object.keys(predicate);
+            const arg = args[0];
+            if (Array.isArray(arg)) {
+                for (let i = 0; i < arg.length; i++) {
+                    const element = arg[i];
+                    any = common.matchesPredicate(element, predicate, predicateKeys);
```
mattolson: Do you think this will be needed by other helpers? If not, it would be better to keep the function contained in this module.

helpers/lib/common.js
```diff
@@ -24,10 +24,20 @@ function isTruthy(val) {
     }
 }
 
+function matchesPredicate(element, predicate, predicateKeys) {
+    for (let j = 0; j < predicateKeys.length; j++) {
```
mattolson: Now that this is not nested, probably better to use `i` for your loop variable

helpers/lib/common.js
```diff
@@ -24,10 +24,20 @@ function isTruthy(val) {
     }
 }
 
+function matchesPredicate(element, predicate, predicateKeys) {
```
mattolson: I would change this so the caller doesn't have to pass in `predicateKeys`. That should be internal to this function.

helpers/any.js
bc-williamkwon: I feel like checking that an object has the same keys as another object can be a useful utility function in itself but there have already been quite a few functions implemented and have only needed it in this instance so maybe better to keep in module.

helpers/lib/common.js
bc-williamkwon: I was thinking that it would be expensive to keep calling Object.keys(predicate) for each element.

bc-williamkwon: But will change it.

helpers/lib/common.js
bc-williamkwon: ok.

bc-williamkwon:
Took all 3 pieces of advice but took advantage of lexical scope and put the function just below const variables in the !common.isEmptyObject block

helpers/any.js
```diff
@@ -15,25 +14,47 @@ const factory = () => {
...
+                for (let i = 0; i < arg.length; i++) {
+                    const element = arg[i];
+                    any = matchesPredicate(element, predicate);
+                    if (any) {
+                        break;
+                    } 
+                }
+            } else if (common.isObject(arg)) {
+                const argIndices = Object.keys(arg);
+                for (let i = 0; i < argIndices.length; i++) {
```
mattolson: I don't think this is the behavior we want. When the first argument is an object, we should be testing that the object has the same properties/values defined in the options hash.

This code appears to be doing that for each attribute of the input object (i.e. assuming a nested object).

Instead, I think you would just do
```
else if (common.isObject(arg)) {
  any = matchesPredicate(arg, predicate);
}
```

bc-williamkwon: I tested the test suite with _.any just now and it matches the current code behavior.

PR #53, opened 2019-06-28
Add configureImage helper

bookernath:
## What? Why?
Add new `configureImage` helper which can be used to control how optimizations are applied to CDN images.

This allows theme developers to selectively disable Akamai Image Manager's features which they may not want interfering with their image delivery in some cases.

Intended HTML usage:

``

This depends on the changes from https://github.com/bigcommerce/paper-handlebars/pull/61, some of which I've pulled into this PR in order to make it work.

## How was it tested?
Unit tests

cc @bigcommerce/storefront-team @chrisboulton @karenwhite @bigcommerce/dev-docs

### DIFF
--- a/helpers/configureImage.js
+++ b/helpers/configureImage.js
@@ -1,0 +1,45 @@
+'use strict';
+const URL = require('url').URL;
+const utils = require('handlebars-utils');
+const common = require('./lib/common.js');
+
+const factory = () => {
+    return function(image, options) {
+        if (!utils.isObject(image) || !utils.isString(image.data) || !common.isValidURL(image.data)) {
+            // Return empty string if this does not appear to be an image object.
+            return ''
+        }
+
+        // 'https://cdn0.bigcommerce.com/s-hash/images/stencil/{:size}/l/some-image.original.jpg?c=2'
+        let url = new URL(image.data);
+        
+        //remove AIM query strings if already set by the application
+        url.searchParams.delete('imbypass');
+        url.searchParams.delete('imdevicesize');
+        url.searchParams.delete('imdeviceformat');
+
+        if (options && utils.isOptions((options))) {
+            if (options.hash['disable-all-optimizations']) {
+                // Add parameter to disable Akamai Image Manager optimization.
+                url.searchParams.set('imbypass', '1');
+                // Query string is now: 'c=2&imbypass=on'
+            } else {
+                if (options.hash['disable-device-sizing']) {
+                    url.searchParams.set('imdevicesize', '0');
+                }
+                if (options.hash['disable-device-formatting']) {
+                    url.searchParams.set('imdeviceformat', '0');
+                }
+            }
+        }
+        // Put special '{:size}' string back (it was URL encoded)
+        image.data = url.toString().replace('%7B:size%7D', '{:size}');
+
+        return image;
+    };
+};
+
+module.exports = [{
+    name: 'configureImage',
+    factory: factory,
+}];
--- a/spec/helpers.js
+++ b/spec/helpers.js
@@ -24,6 +24,7 @@ describe('helper registration', () => {
             'cdn',
             'compare',
             'concat',
+            'configureImage',
             'contains',
             'dynamicComponent',
             'for',
--- a/spec/helpers/configureImage.js
+++ b/spec/helpers/configureImage.js
@@ -1,0 +1,116 @@
+const Lab = require('lab'),
+      lab = exports.lab = Lab.script(),
+      describe = lab.experiment,
+      it = lab.it,
+      testRunner = require('../spec-helpers').testRunner;
+
+describe('configureImage helper', function() {
+    const urlData = 'https://cdn.example.com/path/to/{:size}/image.png?c=2';
+    const urlData_imbypass = 'https://cdn.example.com/path/to/{:size}/image.png?c=2&imbypass=on';
+    const context = {
+        image_url: 'http://example.com/image.png',
+        not_an_image: null,
+        image: {
+            data: urlData
+        },
+        image_imbypass: {
+            data: urlData_imbypass
+        },
+        logoPreset: 'logo',
+    };
+
+    const themeSettings = {
+        logo_image: '600x300',
+    };
+
+    const runTestCases = testRunner({context, themeSettings});
+
+    it('should return blank if a URL is passed', function(done) {
+        runTestCases([
+            {
+                input: '{{getImage (configureImage "http://example.com/image.jpg" disable-all-optimizations=true)}}',
+                output: '',
+            },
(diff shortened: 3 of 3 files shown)


helpers/configureImage.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+const common = require('./lib/common.js');
+
+const factory = globals => {
+    return function(image, mode='') {
```
mattolson: Might be better to use an options hash here, that way people can say
`configureImage image disable-device-formatting=true disable-device-sizing=true` rather than having to support `configureImage image disable-device-sizing-and-formatting` and whatever other permutations might be required.

bookernath: Thanks! I'll look into that.

bookernath:
@mattolson ♻️ 

I still need to add tests, just looking for feedback on approach

Also note that I added another convenience helper

helpers/configureImage.js
```diff
@@ -0,0 +1,48 @@
...
+
+        let params = new URLSearchParams(url.search.slice(1));
+
+        //remove AIM query strings if already set by the application
+        params.delete('imbypass');
+        params.delete('imdevicesize');
+        params.delete('imdeviceformat');
+
+        if (options && common.isOptions((options))) {
+            if (options.get('disable-all-optimizations')) {
```
mattolson: Does `options.get('disable-all-optimizations')` work? I've usually seen this as `options.hash['disable-all-optimizations']`

helpers/configureImage.js
```diff
@@ -0,0 +1,48 @@
...
+                // Add parameter to disable Akamai Image Manager optimization.
+                params.set('imbypass', '1');
+                // Query string is now: 'c=2&imbypass=on'
+            } else {
+                if (options.get('disable-device-sizing')) {
+                    params.set('imdevicesize', '0');
+                } else if (options.get('disable-device-formatting')) {
```
mattolson: You probably want `if`, not `else if` here right?

bookernath: You're right.

helpers/configureImage.js
bookernath: Yep, I was modeling it wrong in my testing. I'll update, thanks for catching.

helpers/configureImage.js
```diff
@@ -0,0 +1,45 @@
...
+        image.data = url.toString().replace('%7B:size%7D', '{:size}');
+
+        return image;
+    };
+};
+
+module.exports = [{
+    name: 'configureImage',
+    factory: factory,
+}];
```
junedkazi: EOL is missing.

helpers/configureImage.js
```diff
@@ -0,0 +1,45 @@
...
+        // 'https://cdn0.bigcommerce.com/s-hash/images/stencil/{:size}/l/some-image.original.jpg?c=2'
+        let url = new URL(image.data);
+        
+        //remove AIM query strings if already set by the application
+        url.searchParams.delete('imbypass');
+        url.searchParams.delete('imdevicesize');
+        url.searchParams.delete('imdeviceformat');
+
+        if (options && utils.isOptions((options))) {
+            if (options.hash['disable-all-optimizations']) {
```
junedkazi: Should we allow the theme developer set these values inside theme config (`globals.getThemeSettings()`) so it can apply to all cdn images ?

helpers/configureImage.js
```diff
@@ -0,0 +1,45 @@
...
+        if (!utils.isObject(image) || !utils.isString(image.data) || !common.isValidURL(image.data)) {
+            // Return empty string if this does not appear to be an image object.
+            return ''
+        }
+
+        // 'https://cdn0.bigcommerce.com/s-hash/images/stencil/{:size}/l/some-image.original.jpg?c=2'
+        let url = new URL(image.data);
+        
+        //remove AIM query strings if already set by the application
+        url.searchParams.delete('imbypass');
```
junedkazi: why are we ignoring them ? If they exists we should respect them and return early.

helpers/lib/common.js
```diff
@@ -0,0 +1,15 @@
...
+        new URL(val);
+        return true;
+    } catch (_) {
+        return false;
+    }
+}
+
+module.exports = {
+    isValidURL,
+};
```
junedkazi: EOL is missing

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,27 @@
...
+        url.searchParams.set(encodeURIComponent(key), encodeURIComponent(value));
+
+        return url.toString();
+    };
+};
+
+module.exports = [{
+    name: 'setURLQueryParam',
+    factory: factory,
+}];
```
junedkazi: EOL is missing.

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,27 @@
...
+        const url = new URL(string);
+
+        url.searchParams.set(encodeURIComponent(key), encodeURIComponent(value));
+
+        return url.toString();
+    };
+};
+
+module.exports = [{
+    name: 'setURLQueryParam',
```
junedkazi: I was wondering if we can move this into its own PR. 🍹

helpers/configureImage.js
bookernath: I was imagining this helper would be used to override the behavior of particular images, not apply broadly across the theme. For example, people want to make certain images nicer for downloading so they might get rid of device formatting (`.webp` is not what you want to download).

helpers/setURLQueryParam.js
bookernath: We could, sure.

helpers/configureImage.js
bookernath: The idea is that we give the theme developer more control. If they want to modify what we return, they're allowed to. They're already using `{{replace}}` for this, I wanted to provide a nicer interface.

helpers/configureImage.js
junedkazi: The reason I ask ... there are a lot stores which have had issues in the past with the way akamai image manager serves some of their images on various devices. Adding the setting at the global level gives them the opportunity to turn it off entirely if there are issue. But I just realized this is going to be a nested helper inside `getImage` so it won't even trigger to respect global settings unless we fetch the image inside `configureImage`.

junedkazi: Can we make it more so that `disable-all-optimizations` and other properties are options on `getImage` so all of this happens behind the scenes.

bookernath: Yep I considered just making `{{getImage}}` more complicated to accommodate this. Perhaps we can do it once we have `{{getImageSrcset}}` and do it all at once.

PR #58, opened 2019-07-01
(WIP) Add "localizePrice" helper

bookernath:
## What? Why?
Adds a new `localizePrice` helper which can be used to automatically configure the display of prices with the magic of [JavaScript's Internationalization API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat).

This helper will override the behavior of store currency settings, and display prices differently according to the provided shopper locale - which can be supplied from browser language header we receive and make available via `request.locale`, or it can be specified statically in the theme. The helper will fall back to the normal stencil "formatted" price if we get any funky input.

The idea is that this will provide "magical" localization capabilities to Stencil, at least where we render pricing.

Since we're just using vanillaJS here, we can perfectly port this method to the frontend as well - for example where we [render PDP pricing from an API call](https://github.com/bigcommerce/cornerstone/blob/master/assets/js/theme/common/product-details.js#L499). So there's an opportunity for parity there.

Depends on:

https://github.com/bigcommerce/bigcommerce/pull/30545
https://github.com/bigcommerce/bigcommerce/pull/30582

## Usage example
----
Assume a price object that looks like this:
```
  "price": {
    "tax_label": "GST",
    "without_tax": {
      "formatted": "$23.00",
      "value": 23,
      "currency": "USD"
    }
  }
```

And a request object that looks like this:

```
  "request": {
    "absolute_path": "\/",
    "host": "ezra-haucks-store.my-staging.zone",
    "is_crawler": false,
    "locale": "en_US",
    "referer": false,
    "user_agent": "Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_14_5) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/75.0.3770.100 Safari\/537.36"
  }
```

Usage would look like this: `{{localizePrice price.without_tax request.locale}}`

And output would be: `$23.00`

But if the request.locale changed to `fr-FR`, or if I manually specified it as such, this would output:

`23,00 $US`

## How was it tested?
----
Looking for feedback on approach before I add tests.

cc @bigcommerce/storefront-team @megdesko @bigcommerce/multi-currency @davidchin

### DIFF
--- a/helpers/localizePrice.js
+++ b/helpers/localizePrice.js
@@ -1,0 +1,46 @@
+'use strict';
+const utils = require('handlebars-utils');
+// Detect ICU support
+const { hasFullICU } = require('./lib/icu-detect.js');
+
+const factory = () => {
+    return function(price, locale) {
+        if (!utils.isObject(price) || !utils.isString(price.currency)
+            || isNaN(price.value) || !utils.isString(price.formatted)) {
+            // Return empty string if this does not appear to be a price object
+            return '';
+        }
+
+        if (!utils.isString(locale) || locale.length < 2) {
+            // Valid browser language strings are at least two characters
+            // https://www.metamodpro.com/browser-language-codes
+            // If provided locale is less than two characters (or not a string),
+            // return the normal formatted price
+            return price.formatted;
+        }
+
+        // If the if full ICU is not installed, fall back to normal price
+        if (!hasFullICU){
+            return price.formatted;
+        }
+
+        // Try to format the price to the provided locale,
+        // but if anything goes wrong,
+        // just return the usual price
+        // Could happen if the full ICU is not installed,
+        // or if an invalid locale is provided.
+        try {
+            return new Intl.NumberFormat(
+                locale, { style: 'currency', currency: price.currency}
+            ).format(price.value);
+        }
+        catch (err){
+            return price.formatted;
+        }
+    };
+};
+
+module.exports = [{
+    name: 'localizePrice',
+    factory: factory,
+}];
--- a/helpers/lib/icu-detect.js
+++ b/helpers/lib/icu-detect.js
@@ -1,0 +1,15 @@
+'use strict';
+
+const hasFullICU = (() => {
+    try {
+        const january = new Date(9e8);
+        const spanish = new Intl.DateTimeFormat('es', { month: 'long' });
+        return spanish.format(january) === 'enero';
+    } catch (err) {
+        return false;
+    }
+})();
+
+module.exports = {
+    hasFullICU
+};
--- a/spec/helpers.js
+++ b/spec/helpers.js
@@ -50,6 +50,7 @@ describe('helper registration', () => {
             'lang',
             'langJson',
             'limit',
+            'localizePrice',
             'money',
             'nl2br',
             'occurrences',
--- a/spec/helpers/localizePrice.js
+++ b/spec/helpers/localizePrice.js
@@ -1,0 +1,38 @@
+const Lab = require('lab'),
+      lab = exports.lab = Lab.script(),
+      describe = lab.experiment,
+      it = lab.it,
+      specHelpers = require('../spec-helpers'),
+      testRunner = specHelpers.testRunner
+
+describe('localizePrice helper', function() {
+    const context = {
+        "price": {
+            "tax_label": "GST",
+            "without_tax": {
+                "currency": "USD",
+                "formatted": "$123,456.78",
+                "value": 123456.78
+            }
+        }
+    };
+
+    const runTestCases = testRunner({context});
+
+    it('should return return correct prices across a number of locales', function(done) {
+        runTestCases([
+            {
+                input: '{{localizePrice price.without_tax "en-US"}}',
+                output: '$123,456.78',
+            },
+            {
+                input: '{{localizePrice price.without_tax "ja-JP"}}',
+                output: '$123,456.78',
+            },
+            {
+                input: '{{localizePrice price.without_tax "de"}}',
+                output: '123.456,78 $',
+            },
+        ], done);
+    });
+});


helpers/localizePrice.js
```diff
@@ -0,0 +1,30 @@
...
+            // Return empty string if this does not appear to be a price object
+            return ''
+        }
+
+        if (!common.isString(locale) || locale.length < 2) {
+            // Valid browser language strings are at least two characters
+            // https://www.metamodpro.com/browser-language-codes
+            // If provided locale is less than two characters (or not a string),
+            // Default to 'en-US'
+            locale = 'en-US';
```
bookernath: An alternate approach here would be to wrap the actual render step in a try/catch and simply fall back to `en-US` if we get any input that JavaScript doesn't recognize as a locale, which might be more resilient.

In Chrome this appears to raise a `RangeError`:

image

Interested in feedback.

filakhtov: I think it is a pretty reasonable approach - try and fallback. Browser know what it supports and we should just rely on that.

helpers/localizePrice.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+const common = require('./lib/common.js');
+
+const factory = globals => {
+    return function(price, locale) {
+
+        if (!common.isObject(price) || !common.isString(price.currency) || isNaN(price.value)) {
+            // Return empty string if this does not appear to be a price object
+            return ''
```
filakhtov: should we return original data without modification here instead?..

bookernath: In this case the assumption is that the theme developer is passing us something incorrect (a completely invalid object, a product image, something else) so we typically "fail" to give the theme developer some clue that they've done something wrong - and protect ourselves against attacks and stuff. That's the pattern in other helpers, but we don't have to follow it here.

I do consider prices to be "mission critical" for rendering a page, so perhaps a behavior that always tries to return something intelligible is better.

@mattolson @junedkazi interested in your perspective.

helpers/localizePrice.js
bookernath: I just realized we can just fall back to `price.formatted` if anything fails - nice clean fallback.

bookernath: Updated to be even more resilient!

helpers/localizePrice.js
```diff
@@ -0,0 +1,38 @@
...
+            // If provided locale is less than two characters (or not a string),
+            // return the normal formatted price
+            return price.formatted;
+        }
+
+        // Try to format the price to the provided locale,
+        // but if anything goes wrong,
+        // just return the usual price
+        try {
+            return new Intl.NumberFormat(
```
davidchin: Is the image of Node used by our storefront renderer built with ICU support? I think only a subset of the full ICU data is supported by default (https://nodejs.org/docs/latest-v10.x/api/intl.html). If it's not, we might need to build with a different image in order to support more than just English locale.

Not sure how to handle the `stencil-cli` as we'll need to enforce a special version of Node on the client. I guess it might be easier if we just use a library instead of `Intl` in order to avoid the environment requirement? 🤔

helpers/localizePrice.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+const common = require('./lib/common.js');
+
+const factory = globals => {
+    return function(price, locale) {
```
davidchin: Does the locale value (coming from Stencil context) takes into consideration of the locale support by the theme? i.e.: The browser could forward an `accept-language` header that the theme could not fulfil. In that case, the price would be localized according to the locale value but not the rest of the page. But I guess it's not really a big issue (as it's understandable by the shopper)?

bookernath: No, it wouldn't take that into account; it's a straight representation of our parsed result of the browser's `Accept-Language` header.

That being said, I don't think that's a bad thing - localizing price display is a different task than localizing all human-readable language, and perhaps we can consider those tasks separately?

helpers/localizePrice.js
bookernath: Super good callout - we'll look into this.

bookernath: https://nodejs.org/docs/latest-v10.x/api/intl.html#intl_embed_the_entire_icu_full_icu

bookernath: @davidchin with the current state of the helper, this should simply fall back to returning `price.formatted` in the case where the proper `Intl` support is not available. I think that's a pretty graceful fallback.

If we go forward with this, we should probably publish instructions on how to set up node with the full ICU data. As long as this is just a helper, that's probably fine. If we start using this helper in Cornerstone, we'll want to make sure all devs have an appropriate environment.

I found this which looks useful: https://www.npmjs.com/package/full-icu

davidchin: Yeah that could work. We can have that as a dependency for the renderer and dev renderer and specify `NODE_ICU_DATA` environment variable when we start the server.

Just a side note, I think `Intl.NumberFormat` won't throw an error if full ICU data is not installed. It would just return the number formatted using English locale. It would only fallback to the number format configured by the merchant if you provide an invalid locale ID. Anyway, it shouldn't be a problem as long as we can get ICU support working. 😄

bookernath: Gotcha - I was going off of what I saw here: https://nodejs.org/docs/latest-v10.x/api/intl.html#intl_detecting_internationalization_support

I was going to add a similar check:

```
const hasFullICU = (() => {
  try {
    const january = new Date(9e8);
    const spanish = new Intl.DateTimeFormat('es', { month: 'long' });
    return spanish.format(january) === 'enero';
  } catch (err) {
    return false;
  }
})();
```

... but it seemed like my existing try/catch was doing something similar.

I could always add this additional check though!

bookernath: Actually now that I look at this with fresh eyes, I see this check is comparing a static result based on what it *would* be if the Intl libraries were properly installed. I'll roll this in...

helpers/localizePrice.js
carsonreinke: 👍 to failing over blind defaults

mattolson: Agree that if a user calls `localizePrice` with an invalid price object, we should fail. In the handlebars helper world, that means returning empty string (for consistency with behavior of other helpers, for example https://github.com/bigcommerce/paper-handlebars/blob/master/helpers/money.js#L25).

mattolson: Need semicolon

helpers/localizePrice.js
```diff
@@ -0,0 +1,55 @@
...
+            // https://www.metamodpro.com/browser-language-codes
+            // If provided locale is less than two characters (or not a string),
+            // return the normal formatted price
+            return price.formatted;
+        }
+
+        // Detect ICU support
+        const hasFullICU = (() => {
+            try {
+                const january = new Date(9e8);
```
mattolson: This is a little confusing. I'd prefer `new Date(0)` with a comment that this is 1/1/1970.

bookernath: This is lifted directly from docs: https://nodejs.org/api/intl.html#intl_detecting_internationalization_support

mattolson:
@bookernath Are we still moving forward with this? I think we are blocked on ICU data availability. Have we made any progress on that and/or tracking that work anywhere?

helpers/localizePrice.js
bookernath: This is now resolved as we have upgraded node versions such that ICU is included by default.

slsriehl:
@bookernath what are the implications here for open ticket [DEVDOCS-1003, handlebars-helpers-reference | Update with localizePrice Helper (Pending Merge)](https://jira.bigcommerce.com/browse/DEVDOCS-1003)?

bookernath:
Hi @slsriehl ,

In general it means we should not document this capability until it's actually live. This PR is sort of just a prototype for now, and might not be live for a while. So it effectively remains "on hold" for now from a documentation perspective. If that's clogging up tubes on your end, feel free to close things as "won't do" for now and we can create a new documentation ticket later when it's needed.

slsriehl:
[DEVDOCS-1003](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-1003)

PR #59, opened 2019-07-02
STRF-6967: Move away from arguments usage

bc-williamkwon:
## What? Why?
https://jira.bigcommerce.com/browse/STRF-6967
Rest parameters are considered more performant than arguments when processing input as an array and call is unacceptable for correct arguments usage.  Other usages such as .length and indexing into valid indices of arguments object are considered ok and not a deoptimization issue.

## How was it tested?
Unit tests still pass.
----

### DIFF
--- a/helpers/all.js
+++ b/helpers/all.js
@@ -3,25 +3,19 @@
 const _ = require('lodash');
 
 /**
- * Yield block only if all arguments are valid
+ * Yield block only if all args are valid
  *
  * @example
  * {{#all items theme_settings.optionA theme_settings.optionB}} ... {{/all}}
  */
 const factory = () => {
-    return function() {
-        var args = [], opts, result;
+    return function(...args) {
 
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
+        // Take the last arg which is a Handlebars options object out of args array
+        const opts = args.pop();
 
-        // Take the last argument (content) out of testing array
-        opts = args.pop();
-
-        // Check if all the arguments are valid / truthy
-        result = _.all(args, function (arg) {
+        // Check if all the args are valid / truthy
+        const result = _.all(args, function (arg) {
             if (_.isArray(arg)) {
                 return !!arg.length;
             }
--- a/helpers/getFontsCollection.js
+++ b/helpers/getFontsCollection.js
@@ -4,8 +4,8 @@ const getFonts = require('./lib/fonts');
 
 const factory = globals => {
     return function() {
-        const options = arguments.length === 1 ? arguments[arguments.length - 1] : null;
-        const fontDisplay = options && options.hash['font-display'] ? options.hash['font-display'] : null;
+        const options = arguments[arguments.length - 1];
+        const fontDisplay = options.hash['font-display'];
         return getFonts('linkElements', globals.getThemeSettings(), globals.handlebars, {fontDisplay});
     };
 };
--- a/helpers/or.js
+++ b/helpers/or.js
@@ -9,21 +9,13 @@ const _ = require('lodash');
  * {{#or 1 0 0 0 0 0}} ... {{/or}}
  */
 const factory = () => {
-    return function() {
-        var args = [],
-            opts,
-            any;
+    return function(...args) {
 
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
+        // Take the last arg which is a Handlebars options object out of args array
+        const opts = args.pop();
 
-        // Take the last argument (content) out of testing array
-        opts = args.pop();
-
-        // Without options hash, we check all the arguments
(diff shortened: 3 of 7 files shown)


helpers/all.js
```diff
@@ -3,25 +3,19 @@
...
 
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
-
-        // Take the last argument (content) out of testing array
-        opts = args.pop();
+        // Take the last arg out of args array
+        const opts = args.pop();
```
bc-zoharmuzafi: @bc-williamkwon why we need to add const?

helpers/getFontsCollection.js
```diff
@@ -4,8 +4,8 @@ const getFonts = require('./lib/fonts');
 
 const factory = globals => {
     return function() {
-        const options = arguments.length === 1 ? arguments[arguments.length - 1] : null;
-        const fontDisplay = options && options.hash['font-display'] ? options.hash['font-display'] : null;
+        const options = arguments[arguments.length - 1];
```
bc-zoharmuzafi: @bc-williamkwon what is ```arguments``` is empty? that might cause undefined? if so, is the next line might fail?

bc-williamkwon: arguments would not be empty because handlebars always generates an options object.

ncheikh: @bc-williamkwon Is there documentation on the handlebars options, or a link to where this is generated in the code?

helpers/all.js
ncheikh: @bc-zoharmuzafi `opts` was listed as a var before at line 13 but since that line was removed it needs to be instantiated.  Before it was reassigned and was thus a var, but as it is never reassigned it can be a const here.

helpers/getFontsCollection.js
bookernath: this is a pattern implemented in many other helpers:

https://github.com/bigcommerce/paper-handlebars/search?q=arguments%5Barguments.length+-+1%5D&unscoped_q=arguments%5Barguments.length+-+1%5D

helpers/all.js
ncheikh: The previous comment said that content was the last argument in the array, is that always the case?

helpers/all.js
```diff
@@ -3,25 +3,19 @@
...
- * Yield block only if all arguments are valid
+ * Yield block only if all args are valid
  *
  * @example
  * {{#all items theme_settings.optionA theme_settings.optionB}} ... {{/all}}
  */
 const factory = () => {
-    return function() {
-        var args = [], opts, result;
+    return function(...args) {
```
ncheikh: Still learning handlebars so this may be an invalid question.  Why do we need to return a function, could we do something like this: `const factory = (...args) => { <logic>  } `

helpers/or.js
```diff
@@ -9,21 +9,12 @@ const _ = require('lodash');
...
-        }
-
-        // Take the last argument (content) out of testing array
-        opts = args.pop();
+        // Take the last arg out of args array
+        const opts = args.pop();
 
-        // Without options hash, we check all the arguments
-        any = _.any(args, function (arg) {
+        const any = _.any(args, function (arg) {
```
ncheikh: Could we add a comment here like the one that is removed to describe what we are doing here?

helpers/getFontsCollection.js
bc-williamkwon: There is a stack overflow topic that gives more insight into handlebars options object https://stackoverflow.com/questions/24606411/what-is-the-options-parameter-in-a-handlebars-helper-function

helpers/all.js
bc-williamkwon: That's something, I didn't look into but don't arrow functions lose their this context?

helpers/or.js
bc-williamkwon: fixed.

helpers/all.js
bc-williamkwon: Yes, what it was popping out of the arguments array was handlebars options object and it still is doing that.

helpers/all.js
bc-williamkwon: I tested the code using that and it can't find all anymore so no we can not do something like that.

helpers/all.js
ncheikh: Could we update the comment to clarify that?  Possibly something similar to the previous comment.  It can be nice to have that additional context, especially when just doing something like an arbitrary .pop()

bc-williamkwon: fixed.

ncheikh: 👍

helpers/all.js
ncheikh: Sorry I didn't follow, could you clarify what happened when you tested it?

bc-williamkwon: Actually, I realized that I had to fix some things due to update of lodash version.  Tests fail with opts.fn is not a function if I change things to be like what was stated previously.

PR #60, opened 2019-07-02
STRF-6968 upgrade lodash

bc-williamkwon:
## What? Why?
Upgrade Lodash due to known security vulnerability in 3.10.1.
https://jira.bigcommerce.com/browse/STRF-6968

An example of an article that mentions minimal performance difference is https://btux1984.wordpress.com/2018/10/06/why-you-should-still-embrace-lodash/
Based on this to gain a performance increase then we can only really do microoptimizations by refactoring everything to a for loop.

A response to https://codeburst.io/why-you-shouldnt-use-lodash-anymore-and-use-pure-javascript-instead-c397df51a66
https://medium.com/@cloud_guy/imo-lodash-is-not-bad-at-all-c8530d9da879

JS fiddle of that response: 
https://jsfiddle.net/o7g30d9r/

image

## How was it tested?
----
The unit tests pass with the replacements of the deprecated functions.

### DIFF
--- a/helpers/all.js
+++ b/helpers/all.js
@@ -15,7 +15,7 @@ const factory = () => {
         const opts = args.pop();
 
         // Check if all the args are valid / truthy
-        const result = _.all(args, function (arg) {
+        const result = _.every(args, function (arg) {
             if (_.isArray(arg)) {
                 return !!arg.length;
             }
--- a/helpers/resourceHints.js
+++ b/helpers/resourceHints.js
@@ -33,7 +33,7 @@ const factory = globals => {
             }
         });
 
-        return new globals.handlebars.SafeString(_.map(hosts, format).join(''));
+        return new globals.handlebars.SafeString(hosts.map(host => format(host)).join(''));
     };
 };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Upgrade Lodash to 4.17.13
 - Reduce usage of arguments
 
 ## 4.0.9
--- a/helpers/any.js
+++ b/helpers/any.js
@@ -17,12 +17,12 @@ const factory = () => {
         const predicate = opts.hash;
 
         if (!_.isEmpty(predicate)) {
-            // With options hash, we check the contents of first arg
-            any = _.any(args[0], predicate);
+            // With options hash, we check the contents of first argument
+            any = _.some(args[0], predicate);
         } else {
             // DEPRECATED: Moved to #or helper
-            // Without options hash, we check all the args
-            any = _.any(args, function (arg) {
+            // Without options hash, we check all the arguments
+            any = _.some(args, function (arg) {
                 if (_.isArray(arg)) {
                     return !!arg.length;
                 }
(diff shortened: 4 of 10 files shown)


bc-zoharmuzafi:
I agree with the upgrading of loadsh in order to resolve the security vulnerability. Deprecating completly the usage of the library involve major change which also effecting the readability of the code. We are still having some performance differences because of usage in loadsh so we might wanna keep it in mind for later refactoring. @mattolson @junedkazi  thoughts?

helpers/all.js
```diff
@@ -9,19 +9,13 @@ const _ = require('lodash');
...
 
-        // Translate arguments to array safely
-        for (var i = 0; i < arguments.length; i++) {
-            args.push(arguments[i]);
-        }
-
-        // Take the last argument (content) out of testing array
-        opts = args.pop();
+        // Take the last arg out of args array
+        const opts = args.pop();
```
bc-zoharmuzafi: why we need to add const?

bc-williamkwon: Just did because my other PR looks like that.

bc-williamkwon:
Yes, I have reviewed the changelog there were a few functions that I had to change as they were removed in favor of the ones that I replaced them with.  There are some behavorial changes but I don't believe that they should be an issue based on our usage.

helpers/all.js
ncheikh: @bc-williamkwon Could you clarify why we need const here at all?

helpers/resourceHints.js
```diff
@@ -33,7 +33,7 @@ const factory = globals => {
             }
         });
 
-        return new globals.handlebars.SafeString(_.map(hosts, format).join(''));
+        return new globals.handlebars.SafeString(_.map(hosts, host => format(host)).join(''));
     };
```
ncheikh: Do we need the lodash map or can use the JS built in map on arrays?

helpers/all.js
bc-williamkwon: This is basically the same file as last.  Const is preferrable to use in es6 and above.  One wouldn't be seeing some parts of the changes in this file if the other pr were to be merged.

helpers/resourceHints.js
bc-williamkwon: We can use the JS built in map but the links listed above show there being no performance difference.  I changed the syntax here as one of the links listed above shows this being faster than using the shorthand string iterator.

ncheikh: I'm not worried about performance of the map function as I assume lodash to be fast and native to be as fast or faster.  I am more looking to reduce the reliance on outside dependencies like lodash where we can.  If we can use more vanilla js it should allow for easier maintenance as we don't have to worry as much about library version changes, security issues and so on.  I am not pushing to remove lodash completely, but where there are simple changes I'm hoping we can make them when supported by native language features.

@junedkazi @mattolson Please let me know if the above approach is not in alignment with the teams approach

I can't speak to the performance gain in regards to latency/first paint as I am still working on my understanding of how these projects are used in regards to a web consumer.  If this becomes part of the bundle to the user there will be performance increases as we will transfer smaller files with less code.

PR #61, opened 2019-07-07
Return image URL as SafeString

bookernath:
## What? Why?

Making getImage return a ` SafeString` instead of a string.

This is because `&` is a special character which gets escaped by Handlebars, which breaks query strings with more than 1 argument in image URLs.

This happens in production today with `imbypass`, example:

https://cdn11.bigcommerce.com/s-grief/images/stencil/1280x1280/products/7295/4490/122901_8k-wallpaper__18797.1558901341.jpg?c=3&amp;imbypass=on

Not that query string should be `?c=3&imbypass=on` but is instead `?c=3&amp;imbypass=on`.

We already have decent checking on this to make sure it's a real image object before we parse it, so this change feels okay to me - but I'm interested to hear from @mattolson on that.

I discovered this while testing https://github.com/bigcommerce/paper-handlebars/pull/53

## How was it tested?

Unit tests on getImage still pass, I added more test cases.

### DIFF
--- a/helpers/getImage.js
+++ b/helpers/getImage.js
@@ -1,6 +1,8 @@
 'use strict';
 
 var _ = require('lodash');
+const SafeString = require('handlebars').SafeString;
+const common = require('./lib/common.js');
 
 const factory = globals => {
     return function(image, presetName, defaultImageUrl) {
@@ -11,7 +13,8 @@ const factory = globals => {
         var width;
         var height;
 
-        if (!_.isPlainObject(image) || !_.isString(image.data) || image.data.indexOf('{:size}') === -1) {
+        if (!_.isPlainObject(image) || !_.isString(image.data)
+            || !common.isValidURL(image.data) || image.data.indexOf('{:size}') === -1) {
             // return empty string if not a valid image object
             defaultImageUrl = defaultImageUrl ? defaultImageUrl : '';
             return _.isString(image) ? image : defaultImageUrl;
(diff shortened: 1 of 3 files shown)


helpers/getImage.js
```diff
@@ -19,8 +20,8 @@ const factory = globals => {
 
         if (_.isPlainObject(presets) && _.isPlainObject(presets[presetName])) {
             // If preset is one of the given presets in _images
-            width = parseInt(presets[presetName].width, 10) || 4096;
-            height = parseInt(presets[presetName].height, 10) || 4096;
+            width = parseInt(presets[presetName].width, 10) || 5120;
+            height = parseInt(presets[presetName].height, 10) || 5120;
```
bookernath: adjusted the maximums to match bcapp's while I was in here.

spec/helpers/getImage.js
```diff
@@ -40,10 +44,6 @@ describe('getImage helper', function() {
                 input: '{{getImage "http://example.com/image.jpg"}}',
                 output: 'http://example.com/image.jpg',
             },
-            {
```
bookernath: removed this as it is duplicative of the test above.

of note, if you pass a URL with 2 query strings, we do escape it - but I think that's kind of an edge case so I don't care about it in this PR.

helpers/getImage.js
```diff
@@ -11,16 +13,17 @@ const factory = globals => {
         var width;
         var height;
 
-        if (!_.isPlainObject(image) || !_.isString(image.data) || image.data.indexOf('{:size}') === -1) {
+        if (!_.isPlainObject(image) || !_.isString(image.data)
+            || !common.isValidURL(image.data) || image.data.indexOf('{:size}') === -1) {
```
bookernath: extra safety by checking for valid URLs.

helpers/getImage.js
junedkazi: Can we add a ticket to seed these from the `siteSettings` context so it always gets the latest value from bcapp.

bookernath: Will do.

PR #62, opened 2019-07-09
Refactor lodash methods to use handlebars-utils instead

bookernath:
## What? Why?
Swapping out lodash methods with handlebars-utils, which is a little nicer for handlebars.

## How was it tested?
Unit tests still pass.

----

### DIFF
--- a/helpers/all.js
+++ b/helpers/all.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const _ = require('lodash');
+const utils = require('handlebars-utils');
 
 /**
  * Yield block only if all args are valid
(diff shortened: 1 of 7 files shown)


bookernath:
In terms of performance, this does seem to run the test suite a few ms faster than master.

ncheikh:
Unless I am mistaken some of these files are the same modified in: https://github.com/bigcommerce/paper-handlebars/pull/59

Just an FYI: @bc-williamkwon

helpers/contains.js
```diff
@@ -12,7 +12,7 @@ const _ = require('lodash');
 const factory = () => {
     return function() {
         var args = Array.prototype.slice.call(arguments, 0, -1),
-            options = _.last(arguments),
+            options = arguments[arguments.length - 1],
```
ncheikh: 👍 appreciate the removal of lodash here for this check

bookernath:
@ncheikh ♻️

PR #64, opened 2019-07-10
STRF-7011 Release 4.0.10

bc-williamkwon:
## What? Why?
Release 4.0.10
- Upgrade Lodash to 4.17.13
- Reduce arguments usage where possible
- Refactor helper functions to use Handlebars utils type checks instead of Lodash type checks
- Add getImageSrcset helper
- Refactor getImage helper to return image URL as SafeString instead of string

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Changelog
 
 ## Unreleased
+
+## 4.1.0
 - Upgrade Lodash to 4.17.13
-- Reduce usage of arguments
+- Reduce arguments usage where possible
+- Refactor helper functions to use Handlebars utils type checks instead of Lodash type checks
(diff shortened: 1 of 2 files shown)


bookernath:
@bc-williamkwon I think this should be 4.1.0 as we've added new features (`getImageSrcset`)

bc-williamkwon:
changed.

bc-williamkwon:
Actually, need to change commit message too.

ncheikh:
What is our release process? (Docs appreciated if they exist)

bc-williamkwon:
The first time, I did it I just made version bumps and updated package-locks to everything for the storefront renderer side.  There were npm publishes that were done but that was not by me.  Not sure where the docs are and need to figure that out.

PR #66, opened 2019-07-24
Add setURLQueryParam helper

bookernath:
## What? Why?
Adds new `setURLQueryParam` helper to help with manipulating query strings on URLs.

## How was it tested?
Unit tests added.

### DIFF
--- a/helpers/setURLQueryParam.js
+++ b/helpers/setURLQueryParam.js
@@ -1,0 +1,37 @@
+'use strict';
+const URL = require('url').URL;
+const utils = require('handlebars-utils');
+const common = require('./lib/common.js');
+const SafeString = require('handlebars').SafeString;
+
+const factory = () => {
+    return function(string, key, value) {
+        if (string instanceof SafeString) {
+            string = string.toString();
+        }
+        if (key instanceof SafeString) {
+            key = key.toString();
+        }
+        if (value instanceof SafeString) {
+            value = value.toString();
+        }
+        if (!utils.isString(string) || !common.isValidURL(string)){
+            throw new TypeError("Invalid URL passed to setURLQueryParam");
+        } else if (!utils.isString(key)){
+            throw new TypeError("Invalid query parameter key passed to setURLQueryParam");
+        } else if(!utils.isString(value)) {
+            throw new TypeError("Invalid query parameter value passed to setURLQueryParam");
+        }
+
+        const url = new URL(string);
+
+        url.searchParams.set(encodeURIComponent(key), encodeURIComponent(value));
+
+        return new SafeString(url.toString());
+    };
+};
+
+module.exports = [{
+    name: 'setURLQueryParam',
+    factory: factory,
+}];
--- a/spec/helpers.js
+++ b/spec/helpers.js
@@ -49,6 +49,7 @@ describe('helper registration', () => {
             'region',
             'replace',
             'resourceHints',
+            'setURLQueryParam',
             'snippet',
             'stripQuerystring',
             'stylesheet',
--- a/spec/helpers/setURLQueryParam.js
+++ b/spec/helpers/setURLQueryParam.js
@@ -1,0 +1,89 @@
+const Lab = require('lab'),
+      lab = exports.lab = Lab.script(),
+      describe = lab.experiment,
+      it = lab.it,
+      specHelpers = require('../spec-helpers'),
+      testRunner = specHelpers.testRunner,
+      renderString = specHelpers.renderString;
+
+
+describe('setURLQueryParam helper', function() {
+    const context = {
+        image_url: 'http://example.com/image.png',
+        not_a_url: null,
+        key: 'referrer',
+        value: 'google',
+    };
+
+    const runTestCases = testRunner({context});
+
+    it('should return a URL with an added parameter given correct input', function(done) {
+        runTestCases([
+            {
+                input: '{{setURLQueryParam "http://example.com/image.jpg" key value}}',
+                output: `http://example.com/image.jpg?${context.key}=${context.value}`,
+            },
+            {
+                input: '{{setURLQueryParam image_url "key" "value"}}',
+                output: `${context.image_url}?key=value`,
+            },
+            {
+                input: '{{setURLQueryParam (setURLQueryParam image_url "key" "value") "key2" "value2"}}',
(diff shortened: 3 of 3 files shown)


junedkazi:
I do see how the helper works. But what are the real world use cases on how this could be used ?

bookernath:
@junedkazi 

- Adding referrer codes to URLs, such as links to category or product pages
- Adding a particular SKU code to a PDP `?sku=abc123` to pre-select options
- Writing handlebars logic to use query strings on API URLs (e.g. `/api/storefront/cart?include=lineItems.physicalItems.options,lineItems.digitalItems.options`)
- Safer way to play with image CDN URLs like `imbypass` (although my other helper is meant to be the go-to for that)

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,31 @@
...
+const utils = require('handlebars-utils');
+const common = require('./lib/common.js');
+const SafeString = require('handlebars').SafeString;
+
+const factory = () => {
+    return function(string, key, value) {
+        if (string instanceof SafeString) {
+            string = string.toString();
+        }
+        if (!utils.isString(string) || !common.isValidURL(string)
```
junedkazi: I think we should break this check into multiple. If `string` is not a string or url we should throw a type error `throw new TypeError("Not a valid url")`

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,31 @@
...
+            || !utils.isString(key) || !utils.isString(value)
+            || !(key.length > 0) || !(value.length > 0)) {
+            // Return empty string if this does not appear to be a URL or if the
+            // key/value are not strings of at least 1 character in length
+            return ''
```
junedkazi: But if they are empty then we should just return the original string. The behavior seems odd to return empty string.

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,31 @@
...
+const common = require('./lib/common.js');
+const SafeString = require('handlebars').SafeString;
+
+const factory = () => {
+    return function(string, key, value) {
+        if (string instanceof SafeString) {
+            string = string.toString();
+        }
+        if (!utils.isString(string) || !common.isValidURL(string)
+            || !utils.isString(key) || !utils.isString(value)
```
junedkazi: Same here if key and value exists and if they are not string throw type error.

bookernath: It seems like in most helpers instead of throwing an error we just return some default output (empty string or pass original value through). Why is it better to throw an error here?

junedkazi: I am suggesting type error here bcoz if the values do exists and if they not string they are breaking a contract defined for the method which is why I recommended throwing an error so it is explicit to the developer that he is using wrong input.

If we do return some value back it creates confusion as to why they are not receiving what they expected.

bookernath: So they'd be able to see this error in CLI, right? What happens in prod, does the whole page fail?

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,28 @@
...
+            // Return empty string if this does not appear to be a URL or if the
+            // key/value are not strings of at least 1 character in length
+            return ''
+        }
+
+        const url = new URL(string);
+
+        url.searchParams.set(encodeURIComponent(key), encodeURIComponent(value));
+
+        return new SafeString(url.toString());
```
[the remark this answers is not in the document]
bookernath: It's necessary if people want to nest this within itself to set more than one query string.

When setting one, it's the different between needing to do `{{{setURLQueryParam}}}` with triple brackets vs being able to do `{{setURLQueryParam}}`.

I think the `isValidURL` check and URL encoding each parameter gives us pretty good safety already. What do you think?

bookernath:
@junedkazi I refactored to throw errors instead of return empty string ♻️

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,35 @@
...
+        }
+        if (!utils.isString(string) || !common.isValidURL(string)){
+            throw new TypeError("Invalid URL passed to setURLQueryParam");
+        }
+        else if (!utils.isString(key)){
+            throw new TypeError("Invalid query parameter key passed to setURLQueryParam");
+        }
+        else if(!utils.isString(value)) {
+            throw new TypeError("Invalid query parameter value passed to setURLQueryParam");
+        } else if (!(key.length > 0) || !(value.length > 0)) {
```
megdesko: 🍹 spacing is different here.

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,35 @@
...
+
+const factory = () => {
+    return function(string, key, value) {
+        if (string instanceof SafeString) {
+            string = string.toString();
+        }
+        if (!utils.isString(string) || !common.isValidURL(string)){
+            throw new TypeError("Invalid URL passed to setURLQueryParam");
+        }
+        else if (!utils.isString(key)){
```
junedkazi: can we move `else if` to the same line as the closing brackets.

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+const URL = require('url').URL;
+const utils = require('handlebars-utils');
+const common = require('./lib/common.js');
+const SafeString = require('handlebars').SafeString;
+
+const factory = () => {
+    return function(string, key, value) {
+        if (string instanceof SafeString) {
```
junedkazi: 👏 nice addition

helpers/setURLQueryParam.js
junedkazi: Not sure if it will ever come here https://github.com/helpers/handlebars-utils/blob/master/index.js#L490

bookernath: good point, I'll remove

helpers/setURLQueryParam.js
```diff
@@ -0,0 +1,31 @@
...
```
junedkazi: I wonder if we should do similar checks for key and value ?

bookernath: makes sense, I'm updating

bookernath:
@junedkazi ♻️

PR #67, opened 2019-07-24
Allow stripQuerystring to accept a SafeString

bookernath:
## What? Why?
When we tried to roll out support for [SafeString on getImage](https://github.com/bigcommerce/paper-handlebars/pull/61) before, we saw issues related to the `getImage` helper being used in conjunction with `stripQuerystring` as a nested expression.

In this PR, I re-write the `stripQuerystring` helper (instead of using the version pulled from [handlebars-helpers](https://github.com/helpers/handlebars-helpers/blob/master/lib/url.js#L106)) to add support for accepting a SafeString as an argument.

I also add tests for problematic cases from the past.

The hope is that this change will allow us to safely roll out that change again, to fix a bug with image URLs.

## How was it tested?
Unit tests added.


bc-williamkwon:
Changes look good to me.

bc-williamkwon:
Actually another thought is that split needs a string.  Would a try and catch work better.  Try to split the string and if that fails catch the error to throw a more meaningful error?

bookernath:
Good point, that's in the original code as well: https://github.com/helpers/handlebars-helpers/blob/master/lib/url.js#L106

I'll update. Must have missed it when copying.

helpers/stripQuerystring.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+const SafeString = require('handlebars').SafeString;
+const utils = require('handlebars-utils');
+
+const factory = () => {
+    return function(url) {
+        if (url instanceof SafeString) {
+            url = url.toString();
+        }
+        if (utils.isString(url)) {
```
junedkazi: Can we change this to a `NOT` case and throw type error and move the return outside.

bookernath: I'm worried about changing the behavior from the existing helper, since this is not a new helper but replacing one that's already in use - if we have a different failure case that might cause weird behavior in themes?

PR #69, opened 2019-07-27
Only generate srcsets up to the dimensions of the image (if known)

bookernath:
## What? Why?
Modify behavior of `getImageSrcset` to only generate srcsets up to the size of the image, *if* we know the image dimensions.

Assumes we'll make updates to pass the dimensions of images into the context in the future.

Today, we could do this for carousel images, which should help optimize home page load speed and reduce inefficiency in the HTML if carousel images less than our maximum sizes are uploaded.

## How was it tested?

Added unit test.

## Todo (maybe in other PRs)

- [ ] Make this work when non-default sizes are used
- [x] Do the same thing to `getImage`

----

### DIFF
--- a/helpers/getImage.js
+++ b/helpers/getImage.js
@@ -24,14 +24,20 @@ const factory = globals => {
             // If preset is one of the given presets in _images
             width = parseInt(presets[presetName].width, 10) || 5120;
             height = parseInt(presets[presetName].height, 10) || 5120;
-            size = width + 'x' + height;
-
+            size = `${width}x${height}`;
         } else if (sizeRegex.test(settings[presetName])) {
             // If preset name is a setting and match the NNNxNNN format
             size = settings[presetName];
+            width = parseInt(size.split('x')[0], 10);
+            height = parseInt(size.split('x')[1], 10);
         } else {
             // Use the original image size
             size = 'original';
+        }
+
+        if (Number.isInteger(image.width) && Number.isInteger(image.height)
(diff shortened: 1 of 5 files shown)


bookernath:
FYI @bigcommerce/dev-docs

helpers/getImageSrcset.js
```diff
@@ -26,16 +26,29 @@ const factory = () => {
...
-                '960w': '960w',
-                '640w': '640w',
-                '320w': '320w',
-                '160w': '160w',
-                '80w': '80w',
-            };
+            if (Number.isInteger(image.width) && Number.isInteger(image.height)){
+                /* If we know the image dimensions, don't generate srcset sizes larger than the image  */
+                srcsets[`${image.width}w`] = `${image.width}w`;
+                const default_srcset_sizes = [2560, 1920, 1280, 960, 640, 320, 160, 80];
```
mattolson: We should use camelCase for variable names

helpers/getImage.js
```diff
@@ -24,16 +24,22 @@ const factory = globals => {
             // If preset is one of the given presets in _images
             width = parseInt(presets[presetName].width, 10) || 5120;
             height = parseInt(presets[presetName].height, 10) || 5120;
-            size = width + 'x' + height;
-
+            size = `${width}x${height}`;
         } else if (sizeRegex.test(settings[presetName])) {
```
ncheikh: 👍 Much prefer interpolation like this

PR #78, opened 2019-10-15
STRF-7500: Ability to set handlebars logger

mattolson:
## What? Why?
Add ability to pass a console-like object to PaperHandlebars, which will then be used for handlebars logging, including for the `log` helper.

## How was it tested?
`npm test`

----


index.js
```diff
@@ -34,8 +34,9 @@ class HandlebarsRenderer {
     * @param {Object} siteSettings - Global site settings, passed to helpers
     * @param {Object} themeSettings - Theme settings (configuration), passed to helpers
     * @param {String} hbVersion - Which version of handlebars to use. One of ['v3', 'v4'] - defaults to 'v3'.
+    * @param {Ojbect} logger - A console-like object to use for logging
```
zackangelo: 🍹 spelling

PR #95, opened 2020-07-26
Add "assignVar", "getVar", "incrementVar", "decrementVar" helpers

bookernath:
## What? Why?
Add `{{assignVar}}` and `{{getVar}}` helpers which allow saving and retrieving custom variables for later use in the template. Similar to `{{inject}` and `{{jsContext}}`, but intended to be used during BE template rendering instead of as a BE->FE API.

Only allowing strings and integers for now.

Also add `{{incrementVar}}` and `{{decrementVar}}` helpers which provide basic counter functionality, incrementing/decrementing the specified variable by 1 each time they are invoked or instantiating them if they don't exist.

## How was it tested?
Unit tests added.

----

cc @bigcommerce/storefront-team @bc-as


mattolson:
Can you describe the use case for this?

bookernath:
@mattolson we've had requests from developers who use Liquid that they really miss this functionality:

https://shopify.github.io/liquid/tags/variable/

I think this sort of custom variable makes it easier to write logic once at the top of a file and simplify logic throughout the rest of the template, because you don't have to do complex nested checks more than once.

helpers/assignVar.js
```diff
@@ -0,0 +1,43 @@
...
+        if (typeof globals.storage.variables === 'undefined') {
+            globals.storage.variables = {};
+        }
+
+        // Make sure the number of total keys is within the limit
+        if (Object.keys(globals.storage.variables).length >= max_keys) {
+            throw new Error(`assignVar helper unique keys may not exceed ${max_keys} in total`);
+        }
+
+        // Store value for later use by jsContext
```
mattolson, line 35: "Later use by `getVar` helper"

bookernath:
@mattolson ♻️

helpers/decrementVar.js
```diff
@@ -0,0 +1,38 @@
...
+            throw new Error("decrementVar helper key must be a string");
+        }
+
+        // Setup storage
+        if (typeof globals.storage.variables === 'undefined') {
+            globals.storage.variables = {};
+        }
+
+        // Make sure the number of total keys is within the limit
+        if (Object.keys(globals.storage.variables).length >= max_keys) {
```
mattolson, line 17: You should probably move this down into the conditional since we only need this guard if creating a new variable.

helpers/decrementVar.js
```diff
@@ -0,0 +1,38 @@
...
+            throw new Error(`Unique keys in variable storage may not exceed ${max_keys} in total`);
+        }
+
+        if (typeof globals.storage.variables[key] !== 'undefined'
+            && Number.isInteger(globals.storage.variables[key])) {
```
mattolson, line 22: You can probably simplify this since `Number.isInteger` will return false if value is undefined. So maybe

```
if (Number.isInteger(globals.storage.variables[key]) {
  // decrement
} else if (typeof globals.storage.variables[key] === 'undefined') {
  // create variable
} else {
  // throw error
}
```

I would also think about changing this so we throw an error if incrementing or decrementing a non-existent value. It's not clear to me that calling `decrementVar` on a nonexistent value should set it to zero.

helpers/getVar.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+const utils = require('handlebars-utils');
+
+const factory = globals => {
+    return function(key) {
+        if (!utils.isString(key)) {
+            throw new Error("getVar helper key must be a string");
+        }
+
+        return globals.storage.variables ? globals.storage.variables[key] : ''
```
mattolson, line 10: Should it be an empty string or undefined?

PR #96, opened 2020-08-24
Add helpers for getting sized images from /content and Image Manager

bookernath:
## What? Why?
Add new helpers for getting sized versions of assets living in WebDav, either in the `/content` folder or in the Image Manager folder (`/product_images/uploaded_images`)

There are helpers for both getting individual image URLs, as well as a `srcset` using the default sizes for the platform.

Unlike the existing `getImage` helpers which accept "stencil image" _objects_, these helpers accept a path to the image in particular object storage folders - so the API is different.

## How was it tested?
Unit tests added.

----


helpers/getContentImage.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+const utils = require('handlebars-utils');
+const common = require('./lib/common');
+const SafeString = require('handlebars').SafeString;
+
+const factory = globals => {
+    return function(path) {
```
junedkazi, line 7: I wonder if we should create multiple helpers. Instead should we just pass `type` as an option which supports pre-defined list like `content` , `image-manager`, `carousel` etc and another flag for `src-set` as a boolean so we can return data into through a single helper. Thoughts ? cc @mattolson

bookernath: these seem like separate enough helpers (that we may even want to evolve independently) that it's nicer to have them separate to me. Just a gut feel.

I could DRY a bit better by making it call the same underlying code with two "modes" for now, though.

bookernath: @junedkazi I've extracted the common code into `lib` to reduce duplication.

bookernath:
I tested this with stencil-cli to confirm everything works 👍

PR #102, opened 2020-09-04
Refactor helpers and remove unneeded dependency

pedelman:
## What? Why?
Refactor our helpers and remove unneeded dependency

## How was it tested?
Unit test coverage. Will be manually testing as well

### DIFF
--- a/helpers/any.js
+++ b/helpers/any.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const utils = require('handlebars-utils');
 
 /**
@@ -15,15 +14,15 @@ const factory = () => {
         let any;
         // Take the last arg which is a Handlebars options object out of args array
         const opts = args.pop();
+        const collection = args[0];
         const predicate = opts.hash;
 
         if (!utils.isEmpty(predicate)) {
-            // With options hash, we check the contents of first argument
-            any = _.some(args[0], predicate);
+            any = Object.entries(predicate).some(([key, value]) => collection.some(arg => arg[key] === value));
         } else {
             // DEPRECATED: Moved to #or helper
             // Without options hash, we check all the arguments
-            any = _.some(args, function (arg) {
+            any = args.some(arg => {
                 if (utils.isArray(arg)) {
                     return !!arg.length;
                 }
--- a/helpers/deprecated/pick.js
+++ b/helpers/deprecated/pick.js
@@ -1,13 +1,23 @@
 'use strict';
 
-const _ = require('lodash');
-
 const factory = () => {
     /**
      * @deprecate
      */
-    return function() {
-        return _.pick.apply(null, arguments);
+    return function(obj, ...args) {
+        return args.reduce(
+            (
+                pickedElements,
+                element
+            ) => {
+                if (obj && obj.hasOwnProperty(element)) {
+                    pickedElements[element] = obj[element];
+                }
+
+                return pickedElements;
+            },
+            {}
+        );
     };
 };
 
--- a/helpers/lib/fonts.js
+++ b/helpers/lib/fonts.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
-
 const fontProviders = {
     'Google': {
         /**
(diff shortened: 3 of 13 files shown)


helpers/lib/fonts.js
```diff
@@ -21,30 +19,30 @@ const fontProviders = {
...
                     return;
                 }
 
-                if (_.isUndefined(weights)) {
+                if (typeof weights === 'undefined') {
                     weights = '';
                 }
 
-                if (!_.isArray(familyHash[familyKey])) {
+                if (typeof familyHash[familyKey] === 'undefined') {
```
mattolson, line 35: There is also `Array.isArray` for this

junedkazi, line 35:
```suggestion
                if (!Array.isArray(familyHash[familyKey])) {
```

helpers/lib/fonts.js
```diff
@@ -84,58 +82,90 @@ const fontProviders = {
...
-        });
-        return fontLoaderConfig;
+        return Object.entries(parsedFonts).reduce(
+            (
+                configs,
+                [key, value]
+            ) => {
+                configs[key] = fontProviders[key].buildFontLoaderConfig(value);
+
+                return Object.entries(configs).reduce(
```
jairo-bc, line 154: Can't be it simplified to 
`
return {...configs, ...fontProviders[key].buildFontLoaderConfig(value)};
`?

helpers/any.js
```diff
@@ -15,15 +14,15 @@ const factory = () => {
         let any;
         // Take the last arg which is a Handlebars options object out of args array
         const opts = args.pop();
+        const collection = args[0];
         const predicate = opts.hash;
 
         if (!utils.isEmpty(predicate)) {
-            // With options hash, we check the contents of first argument
-            any = _.some(args[0], predicate);
+            any = Object.entries(predicate).some(([key, value]) => collection.some((arg) => arg[key] === value));
```
junedkazi, line 21: don't need the `()` around `arg`

helpers/lib/fonts.js
pedelman: Eslint complain on this sadly. Worth investigating the reason in future PR definitely.

```
/Users/patrick.edelman/dev/paper-handlebars/helpers/lib/fonts.js
  154:31  error  Parsing error: Unexpected token ..
```

helpers/lib/fonts.js
```diff
@@ -84,58 +82,90 @@ const fontProviders = {
...
+                return collectedFonts;
+            },
+            {}
+        );
 
     // Parse font strings based on provider
-    const parsedFonts = _.mapValues(collectedFonts, function(value, key) {
-        return fontProviders[key].parser(value);
-    });
+    const parsedFonts = Object.entries(collectedFonts).reduce(
```
junedkazi, line 116: It would be nice we can move this into a helper method with a callback to avoid duplication

junedkazi (dismissed):

helpers/deprecated/pick.js
```diff
@@ -1,13 +1,23 @@
...
      */
-    return function() {
-        return _.pick.apply(null, arguments);
+    return function(obj, ...args) {
+        return args.reduce(
+            (
+                pickedElements,
+                element
+            ) => {
+                if (obj.hasOwnProperty(element)) {
```
junedkazi, line 13:
```suggestion
                if (obj && obj.hasOwnProperty(element)) {
```

helpers/deprecated/pick.js
```diff
@@ -1,13 +1,23 @@
...
 const factory = () => {
     /**
      * @deprecate
      */
-    return function() {
-        return _.pick.apply(null, arguments);
+    return function(obj, ...args) {
+        return args.reduce(
+            (
+                pickedElements,
```
junedkazi, line 10: can we just pass `obj` here ? and then at the bottom we do 
```
    obj[element] = obj[element];
```

pedelman: I think its probably better to avoid shadowing. Its not that we would be passing obj here, its what we would be naming the accumulator.

helpers/lib/fonts.js
pedelman: Agree, I prefer to avoid that level of refactor in this PR if thats ok with you, maybe we can do in future PR?

junedkazi (approved):

junedkazi (approved):

helpers/lib/fonts.js
MaxGenash: @pedelman, @jairo-bc the reason is that we have specified to support ES6 only in our .eslintrc, just need to update the config to support the newer JS syntax. I had a similar issue with Stencil-CLI previously.
Created a PR with the updated ESLint: https://github.com/bigcommerce/paper-handlebars/pull/103

PR #103, opened 2020-09-07
feat: update ESLint to support newer JS syntax and drop Node 8

MaxGenash:
## What? Why?

- Updated .eslintrc to support newer JS syntax (set es2018 which is fully supported by node 10).
- Updated the ESLInt to the last version.
- Dropped Node 8 since we don't support it here

## How was it tested?

`npm run linter`


jairo-bc (approved):

mattolson (dismissed):

mattolson:
I don't believe we support Node 8.x anymore since it is EOL. Is that correct @junedkazi?

.eslintrc
```diff
@@ -1,14 +1,13 @@
...
     "sourceType": "module",
     "ecmaFeatures": {
       "impliedStrict": true
-    },
+    }
   },
   "env": {
     "node": true,
-    "es6": true,
+    "es2020": true
```
mattolson, line 10: es2020 is still not supported in Node 10. I hesitate to add linting support since it might lead to us introducing features that depend on newer JS that is not yet supported in our runtime.

https://node.green/#ES2020

junedkazi: so we have to hold off until we deprecate stapler ?

mattolson: Until we upgrade Node to a version that supports it. Looks like Node 12 has decent support.

MaxGenash: @mattolson, @junedkazi 
Actually, Node will just break when encounter unsupported syntax, so if we have good test coverage - we won’t miss these parts.
But I can just set "es2018" for the linter, it seems to have good support by Node 10

MaxGenash:
> I don't believe we support Node 8.x anymore since it is EOL. Is that correct @junedkazi?

I thought we support Node 8.x because we test on it in Travis

jairo-bc:
We might need to change `linter` to `npm run lint` to have consistency with all packages

junedkazi:
We support node 8 for paper 2.x branch since stapler still runs on node 8. We don't need to support it for paper handlebars since it is used by storefront renderer which is on node 10. So we can safely drop node 8 support here.

MaxGenash:
> We support node 8 for paper 2.x branch since stapler still runs on node 8. We don't need to support it for paper handlebars since it is used by storefront renderer which is on node 10. So we can safely drop node 8 support here.

Ok, cool, thanks!
So I'll remove Node 8 from Travis in this PR

MaxGenash:
> I don't believe we support Node 8.x anymore since it is EOL. Is that correct @junedkazi?

Ok, dropped node 8 from here

MaxGenash:
> We might need to change `linter` to `npm run lint` to have consistency with all packages

updated

junedkazi (approved):

PR #105, opened 2020-09-10
STRF-8666: Adds test coverage for contains helper

pedelman:
## What? Why?
Adds test coverage for contains helper. It looks like we were previously missing coverage for this helper. Added some additional edge case testing.

## How was it tested?
Unit tests added.

----

### DIFF
--- a/helpers/contains.js
+++ b/helpers/contains.js
@@ -1,7 +1,7 @@
 'use strict';
 
 /**
- * Is any value included in a collection or a string?
+ * Is any value included in a collection or a string or top level value in an object
  *
  * @example
  * {{#contains fonts "Roboto"}} ... {{/contains}}
@@ -10,7 +10,7 @@
 const factory = () => {
     return function(container, value) {
         const options = arguments[arguments.length - 1];
-        const contained = container.includes(value);
+        const contained = containerIncludesValue(container, value);
 
         // Yield block if true
         if (contained) {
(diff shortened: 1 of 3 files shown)


junedkazi (approved):

helpers/contains.js
```diff
@@ -21,6 +21,28 @@ const factory = () => {
...
+        case 'string':
+            return container.includes(value);
+        case 'object':
+            if (container) {
+                if  (Array.isArray(container)) {
+                    return container.includes(value);
+                }
+
+                // Check if the supplied value matches a top level value in the container object
+                return Object.values(container).some(objValue => {
```
junedkazi, line 35: I would have liked for us to keep this simple like `Object.values(object1).includes(value);` but I guess if lodash is supporting complex type this make is super difficult for us

helpers/contains.js
```diff
@@ -10,7 +10,7 @@
 const factory = () => {
     return function(container, value) {
         const options = arguments[arguments.length - 1];
-        const contained = container.includes(value);
+        const contained = containerIncludesValue(container, value);
```
jairo-bc, line 13: What is the general reason to support more data types?

pedelman: I wanted to make sure to preserve as much backward compatibility as possible. Lodash supported more options so its not as simple of a refactor as I initially thought, we also didnt explicitly test this case, but our public helper docs point out object as a possible value.

PR #106, opened 2020-09-10
Update handlebar helpers to use contains and pluck and drop native support for them

junedkazi:
## What? Why?
Update handlebar helpers to use contains and pluck and drop native support for them

## How was it tested?
working on it
----


package.json
```diff
@@ -27,7 +27,7 @@
   "dependencies": {
     "@bigcommerce/handlebars-v4": "4.7.6",
     "handlebars": "3.0.8",
-    "handlebars-helpers": "0.8.4",
+    "handlebars-helpers": "0.10.0",
```
mattolson, line 30: handlebars-helpers doesn't follow semver and introduced some breaking changes -- check their changelog

junedkazi: It is mostly addition. I am going through the ones which were changed to confirm we are not using them

spec/helpers.js
```diff
@@ -167,6 +166,7 @@ describe('helper registration', () => {
             'titleize',
             'trim',
             'uppercase',
+            'ellipsis',
```
junedkazi, line 169: So this will be a breaking change. Earlier if html tag was present it used to ignore/sanitize it before running `ellipsis` on the actual test

jairo-bc (approved):

spec/helpers/thirdParty.js
```diff
@@ -73,6 +97,245 @@ describe('third party handlebars-helpers', function() {
...
+                runTestCases([
+                    {
+                        input: '{{#contains haystack needle}}FOUND{{/contains}}',
+                        output: 'FOUND',
+                        context: {
+                            needle: "hello",
+                            haystack: "hello world"
+                        },
+                    },
+                    // {
```
MaxGenash, line 111: Is it a temporary block of commented code that should be removed, or there is some special sense in it?

junedkazi: These test are failing bcoz even though the helper says it supports object it doesn't

PR #107, opened 2020-09-10
STRF-8667: Adds test coverage for any helper

pedelman:
## What? Why?
Adds test coverage for any helper in support of dropping lodash.

## How was it tested?
Unit tests added.

----

### DIFF
--- a/spec/helpers/any.js
+++ b/spec/helpers/any.js
@@ -44,6 +44,27 @@ describe('any helper (with option hash)', function() {
                 input: '{{#any arrayWithObjs num=2}}{{big}}{{/any}}',
                 output: '',
             },
+            {
+                input: '{{#any foobar num=2}}{{big}}{{/any}}',
+                output: '',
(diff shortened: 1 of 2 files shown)


junedkazi (approved):

spec/helpers/any.js
```diff
@@ -44,6 +44,27 @@ describe('any helper (with option hash)', function() {
...
+            },
+            {
+                input: '{{#any foobar num=2}}{{big}}{{/any}}',
+                output: '',
+                context: {
+                    foobar: undefined
+                }
+            },
+            {
+                input: '{{#any undefined num=1}}{{big}}{{/any}}',
```
mattolson, line 62: If you're trying to test for `undefined`, you probably want `{{#any num=1}}{{big}}{{/any}}`

pedelman: Will add this in #105

PR #108, opened 2020-09-11
Fix pluck helper and support nested fetch

junedkazi:
## What? Why?
Fix pluck helper and support nested fetch

## How was it tested?
unit test in place
----

### DIFF
--- a/helpers/pluck.js
+++ b/helpers/pluck.js
@@ -1,8 +1,29 @@
 'use strict';
 
+const helpers = require('handlebars-helpers');
+
 const factory = () => {
     return function(collection, path) {
-        return collection.map(item => item[path]);
+        if (typeof path !== 'string') {
+            return '';
+        }
+
+        if (Array.isArray(collection)) {
+            let res = [];
+            for (let i = 0; i < collection.length; i++) {
+                const val = helpers.utils.get(collection[i], path);
+                if (typeof val !== 'undefined') {
+                    res.push(val);
+                }
+            }
+            return res;
+        }
+
+        if (typeof collection === 'object') {
+            const result = helpers.utils.get(collection, path);
+            return result ? result : '';
+        }
+        return '';
     };
 };
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bigcommerce/stencil-paper-handlebars",
-  "version": "4.5.0-rc.1",
+  "version": "4.5.0-rc.2",
   "description": "A paper plugin to render pages using Handlebars.js",
   "main": "index.js",
   "author": "Bigcommerce",
--- a/spec/helpers/pluck.js
+++ b/spec/helpers/pluck.js
@@ -7,9 +7,10 @@ const Lab = require('lab'),
 describe('pluck helper', function() {
     const context = {
         users: [
-          { 'user': 'barney', 'age': 36 },
-          { 'user': 'fred',   'age': 40 }
-        ]
+          { 'user': 'barney', 'age': 36, 'image': { 'url': 'barney.jpg' } },
+          { 'user': 'fred',   'age': 40, 'image': { 'url': 'fred.jpg' }  }
+        ],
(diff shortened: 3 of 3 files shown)


helpers/pluck.js
```diff
@@ -1,8 +1,21 @@
...
+const helpers = require('handlebars-helpers');
+
 const factory = () => {
     return function(collection, path) {
-        return collection.map(item => item[path]);
+        if (collection === null || typeof collection === 'undefined' || typeof collection !== 'object') {
+            return '';
+        }
+
+        let res = [];
```
junedkazi, line 11: picked up this logic from here https://github.com/helpers/handlebars-helpers/blob/master/lib/array.js#L445

helpers/pluck.js
```diff
@@ -1,8 +1,21 @@
...
+        for (let i = 0; i < collection.length; i++) {
+            const val = helpers.utils.get(collection[i], path);
```
junedkazi, line 13: https://github.com/helpers/handlebars-helpers/blob/master/lib/utils/utils.js#L30

helpers/pluck.js
MaxGenash: @junedkazi So, why we don’t just grab "pluck" from handlebars-helpers? (and similarly, we can do with other utils)

helpers/pluck.js
```diff
@@ -1,8 +1,21 @@
 'use strict';
 
+const helpers = require('handlebars-helpers');
+
 const factory = () => {
     return function(collection, path) {
-        return collection.map(item => item[path]);
+        if (collection === null || typeof collection === 'undefined' || typeof collection !== 'object') {
```
mattolson, line 7: Is `pluck` supposed to operate on arrays or objects?

junedkazi: As per our documentation both

helpers/pluck.js
```diff
@@ -1,8 +1,21 @@
...
+            return '';
+        }
+
+        let res = [];
+        for (let i = 0; i < collection.length; i++) {
```
mattolson, line 12: Can you iterate through objects like this? I've never done it that way. Can you add a test case?

mattolson (approved):

helpers/pluck.js
```diff
@@ -1,8 +1,30 @@
 'use strict';
 
+const helpers = require('handlebars-helpers');
+const object_helper = helpers.object();
```
mattolson, line 4: This helper is so simple that it's probably not worth bringing in this dependency.

helpers/pluck.js
```diff
@@ -1,8 +1,30 @@
...
+
 const factory = () => {
     return function(collection, path) {
-        return collection.map(item => item[path]);
+        if (typeof collection !== 'object' || typeof path !== 'string') {
```
mattolson, line 8: Perhaps this whole function could be simplified:

```
if (Array.isArray(collection)) {
  // do array stuff
  return something;
}

if (typeof collection === 'object') {
 // do object stuff
 return something;
}

return '';
```

helpers/pluck.js
```diff
@@ -1,8 +1,30 @@
...
+            return '';
+        }
+
+        if (!Array.isArray(collection) && object_helper.isObject(collection)) {
+            const result = object_helper.get(path, collection);
+            console.log("checking object inside", collection, path, result);
```
mattolson, line 14: You should remove log statements like this prior to merge.

helpers/pluck.js
```diff
@@ -1,8 +1,27 @@
 'use strict';
 
+const helpers = require('handlebars-helpers');
+
 const factory = () => {
     return function(collection, path) {
-        return collection.map(item => item[path]);
+        const isValidPath = typeof path === 'string';
```
mattolson, line 7: I would just return early for this rather than duplicating this condition multiple times below.

```
if (typeof path !== 'string') {
  return '';
}
```

jairo-bc (approved):

mattolson (approved):

PR #109, opened 2020-09-23
PEN-1939 refactor date helper

jairo-bc:
## What? Why?
 - Moved date and moment helpers to the codebase. 
 - date.js is not opened for the world
 - Upgraded moment.js version

## How was it tested?

using **stencil** and **npm test** 


helpers/date.js
```diff
@@ -0,0 +1,63 @@
...
+        if (utils.isString(str) && !pattern) {
+          return moment().format(str);
+        }
+      
+        return moment(str).format(pattern);
+      };
+};
+
+module.exports = [{
+    name: 'date',
```
mattolson, line 58: I don't think we need to expose this as `date`. We currently only expose `moment`.

spec/helpers/date.js
```diff
@@ -0,0 +1,40 @@
+const Lab = require('lab'),
+      lab = exports.lab = Lab.script(),
+      describe = lab.experiment,
+      it = lab.it,
+      testRunner = require('../spec-helpers').testRunner;
+
+
+describe('date/moment helper', function() {
+  const context = { date: '01/20/2019' };
```
mattolson, line 9: Just to make the tests less confusing (because we have both a context variable and a helper named `date`) can you rename this to something other than `date`? Maybe `today` or something?

spec/helpers/date.js
```diff
@@ -0,0 +1,40 @@
...
+
+  const runTestCases = testRunner({context});
+  const addZeroToDate = (date) => ('0' + date).slice(-2);
+
+  it('should return a default formatted moment date', function(done) {
+    const now = new Date();
+    console.log(`${addZeroToDate(now.getMonth() + 1)} ${addZeroToDate(now.getDate())}`);
```
mattolson, line 16: We shouldn't include console logging in tests

spec/helpers/thirdParty.js
```diff
@@ -77,22 +77,6 @@ describe('third party handlebars-helpers', function() {
...
     });
 
-    describe('date helpers', function() {
-
-        describe('contains moment', function() {
-            it('renders the date in the format specified', function(done) {
-                const now = new Date();
-                runTestCases([
-                    {
-                        input: `{{#moment "1 year ago" "YYYY"}}{{/moment}}`,
```
mattolson, line 87: Are we removing this functionality? If not, can you replicate this test case?

helpers/date.js
jairo-bc: You are right, current handlebars-helpers version doesn't expose `date`

spec/helpers/date.js
jairo-bc: So, there will be only `moment` helper, so `date` shouldn't be confusing

spec/helpers/date.js
jairo-bc: removed

spec/helpers/thirdParty.js
jairo-bc: removing

MaxGenash (approved):

PR #110, opened 2020-10-06
STRF-8585 Escape injected values

bdav87:
## What? 
Escaping values passed into jsContext through the inject helper using Handlebars escapeExpression.

## Why?
Currently, users can enter values on the account page that aren't escaped, potentially opening up site visitors to XSS attacks when the values are injected into jsContext.

By escaping the values as they are injected, any unsafe values (i.e. script tags) will be rendered as plain text on the page, even if a developer is feeding them into the page with innerHTML.

## How was it tested?
Used my fork of paper-handlebars within a locally referenced version of paper with storefront-renderer running in VM. I uploaded a customized version of the latest Cornerstone to my VM store with extra account values injected on the page so I could see them in the browser inspect element console.

Injected fields:
image

Example of account page with unsafe characters:
image

Example of customer address with unsafe characters:
image

Output of jsContext:
image

----

cc @junedkazi

### DIFF
--- a/helpers/inject.js
+++ b/helpers/inject.js
@@ -9,8 +9,13 @@ const factory = globals => {
             if (typeof value === 'string') {
                 result = globals.handlebars.escapeExpression(value);
             }
-            if (typeof value === 'object' && value !== null) {
+            if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
                 result = filterObjectValues(value);
+            }
+            if (Array.isArray(value)) {
+                result = value.map(item => {
+                    return filterValues(item);
+                });
             }
         }
         return result;
@@ -18,16 +23,7 @@ const factory = globals => {
     function filterObjectValues(obj) {
         let filteredObject = {};
         Object.keys(obj).forEach(key => {
-            if (typeof obj[key] === 'string') {
-                filteredObject[key] = globals.handlebars.escapeExpression(obj[key]);
-                return;
-            }
-            if (typeof obj[key] === 'object' && obj[key] !== null) {
-                filteredObject[key] = filterObjectValues(obj[key]);
-                return;
-            } else {
-                filteredObject[key] = obj[key];
-            }
+            filteredObject[key] = filterValues(obj[key]);
         });
         return filteredObject;
     }
--- a/helpers/inject.js
+++ b/helpers/inject.js
@@ -1,6 +1,37 @@
 'use strict';
 
 const factory = globals => {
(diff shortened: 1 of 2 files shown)


bookernath:
@junedkazi @mattolson how likely is this to be a breaking change to existing themes?

helpers/inject.js
```diff
@@ -1,6 +1,37 @@
 'use strict';
 
 const factory = globals => {
+    function filterValues(value) {
+        let result = value;
+        try {
+            JSON.parse(value);
```
junedkazi, line 7: Can we make sure we have test cases for all scenarios

helpers/inject.js
```diff
@@ -1,6 +1,37 @@
...
+        Object.keys(obj).forEach(key => {
+            if (typeof obj[key] === 'string') {
+                filteredObject[key] = globals.handlebars.escapeExpression(obj[key]);
+                return;
+            }
+            if (typeof obj[key] === 'object' && obj[key] !== null) {
+                filteredObject[key] = filterObjectValues(obj[key]);
+                return;
+            } else {
+                filteredObject[key] = obj[key];
```
junedkazi, line 29: should there be an else case ? and should we be adding the data as filtered without verifying it ?

junedkazi (commented):
Overall I am fine with the approach. I think this will address XSS across all stores. 
Adding @bigcommerce/security for visibility.

helpers/inject.js
bdav87: The else might not be necessary, since there's a return on the previous 2 conditions, so I could just assign the value to the filtered object.

I'm not sure what else I'd be verifying, since other types should be fine to include without worrying about XSS. The purpose of the filterObjectValues function is to make sure any strings that can be injected are escaped, even if they are in nested objects/arrays.

helpers/inject.js
bdav87: Building the test cases is a good idea, because while writing tests I found that I need to update my changes to account for Array specifically (we have some injectable values that are arrays nested in objects nested in arrays etc. like blogs).  😅

junedkazi: I think it should be caught by the object check in the catch

bc-benjaminpogue (approved):
This looks good from the security side. The only issue I potentially see here is that we may end up double escaping some strings where we have escaped them prior. @junedkazi I am not sure how much of an issue that would be, but thought I would call that out.

PR #112, opened 2020-11-05
Support "classes" argument on getFontLoaderConfig

bookernath:
## How was it tested?

----


helpers/lib/fonts.js
```diff
@@ -144,7 +144,7 @@ module.exports = function(format, themeSettings, handlebars, options) {
 
     case 'webFontLoaderConfig':
         // Build configs
-        return Object.entries(parsedFonts).reduce(
+        return Object.assign(Object.entries(parsedFonts).reduce(
```
jairo-bc, line 147: Can we use spread operator over here?

PR #134, opened 2021-03-23
STRF-8093 Fix handlebars helpers using node package url

jairo-bc:
## What? Why?

Add additional check for running in NodeJS environment, because the package is a global variable on browser-like environment 

## How was it tested?

`npm test `
----


helpers/setURLQueryParam.js
```diff
@@ -17,7 +17,7 @@ const factory = () => {
             throw new TypeError("Invalid query parameter value passed to setURLQueryParam");
         }
 
-        const url = new URL(string);
+        const url = NodeJsURL ? new NodeJsURL(string) : new URL(string);
```
MaxGenash, line 20: I'd move out this check to a separate file and reuse kinda `new UrlParser()`

helpers/lib/common.js
```diff
@@ -1,10 +1,14 @@
...
+const NodeJsURL = require('url').URL;
 const SafeString = require('handlebars').SafeString;
 
 function isValidURL(val) {
     try {
-        new URL(val);
+        if (NodeJsURL) {
+            new NodeJsURL(val);
+        } else {
+            new URL(val);
```
junedkazi, line 10: What is the difference between the 2 methods. If vanilla JS method provides the same capabilities then do we need to do this node js check ? can't we simply just switch over to the JS version.

jairo-bc: You are right. I thought NodeJs supports only package requiring.

helpers/setURLQueryParam.js
jairo-bc: I've just removed totally that requiring, so we don't need to introduce a package

MaxGenash (approved):

junedkazi (approved):
Can you please make sure we have some tests in place to confirm there is no breaking behavior from before and after.

PR #138, opened 2021-04-05
BCTHEME-276

bdav87:
## What? 
Providing the option to escape values in the inject helper.

## Why?
To help theme developers avoid XSS vulnerabilities when injecting customer supplied data like account fields. When passing true to the inject helper, any values with special characters will be escaped.

The default value is false, to have backwards compatibility with existing themes. This will let devs opt in to escaping injected values.

## How was it tested?
- Created unit tests
- Ran storefront-renderer locally to compare behavior when the 'escape' value for inject is true, false, and not provided

set to false or no argument provided:
image

set to true:
image

----

### DIFF
--- a/helpers/inject.js
+++ b/helpers/inject.js
@@ -1,7 +1,34 @@
 'use strict';
 
 const factory = globals => {
-    return function(key, value) {
+    function filterValues(value) {
+        let result = value;
+        try {
+            JSON.parse(value);
+        } catch (e) {
+            if (typeof value === 'string') {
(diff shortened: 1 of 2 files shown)


jairo-bc:
Maybe it's better to introduce `safeInject` for the user input? 
What are your thoughts? @junedkazi

briansilvey:
@junedkazi would you be able to take a look at the changes that @bdav87 has made after y'alls last meeting?

jairo-bc:
Please reopen if needed

PR #152, opened 2021-11-16
Support floats for assignVar/getVar

bookernath:
## What? Why?
Add support for floats in assignVar/getVar helpers by customer request.

## How was it tested?
Test added.

----

### DIFF
--- a/helpers/assignVar.js
+++ b/helpers/assignVar.js
@@ -11,9 +11,9 @@ const factory = globals => {
             throw new Error("assignVar helper key must be a string");
         }
 
-        // Validate that value is a string or integer
-        if (!utils.isString(value) && !Number.isInteger(value)) {
-            throw new Error("assignVar helper value must be a string or a number (integer)");
+        // Validate that value is a string or Number (int/float)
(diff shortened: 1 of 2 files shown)


jairo-bc (approved):

helpers/assignVar.js
```diff
@@ -11,9 +11,9 @@ const factory = globals => {
             throw new Error("assignVar helper key must be a string");
         }
 
-        // Validate that value is a string or integer
-        if (!utils.isString(value) && !Number.isInteger(value)) {
-            throw new Error("assignVar helper value must be a string or a number (integer)");
+        // Validate that value is a string or Number (int/float)
+        if (!utils.isString(value) && isNaN(value)) {
```
mattolson, line 15: I would use `Number.isFinite(value)` here for a more robust and less surprising implementation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite

bookernath:
@jairo-bc @mattolson ♻️ 

I think this is good to merge.

jmwiese (approved):

PR #156, opened 2022-02-14
String replace helper with more options

thearcticalex:
## What? Why?

This is an inline string replacement helper that gives the user more control in how many occurrences of the substring to be replaced. 

## How was it tested?

It was tested by giving several test cases including inputs that are out of bound. 

`{{strReplace string substr newSubstr}}` would replace all substring by default

`{{strReplace string substr newSubstr count}}` would replace count of occurrences.

Invalid string input would return invalid. 
Invalid count would return original string and extra count (count > occurrences) would regard as replace all. 

----

### DIFF
--- a/helpers/strReplace.js
+++ b/helpers/strReplace.js
@@ -1,22 +1,40 @@
 'use strict';
 
 const factory = () => {
-    return function(string, substr, newSubstr, token) {
-        if (typeof string !== 'string' || typeof substr !== 'string' || typeof newSubstr !== 'string') {
+    return function(str, substr, newSubstr, iteration) {
+        if (typeof str !== 'string' || typeof substr !== 'string' || typeof newSubstr !== 'string') {
             return 'Invalid Input';
         }
 
-        if (token && typeof token === 'string') {
-            return string.replace(new RegExp(escapeRegex(substr), token), newSubstr);
-        } else {
-            return string.replace(new RegExp(escapeRegex(substr), 'g'), newSubstr);
+        if (typeof iteration !== 'number') {
+            return str.replace(new RegExp(escapeRegex(substr), 'g'), newSubstr);
         }
 
+        const occurrence = getOccurrences(str, substr);
+
+        if (iteration > 0 && occurrence > 0) {
+            if (iteration >= occurrence) {
+                return str.replace(new RegExp(escapeRegex(substr), 'g'), newSubstr);
+            } else {
+                let result = str;
+                for (let i = 0; i < iteration; i++) {
+                    result = result.replace(substr, newSubstr);
+                }
+                return result;
+            }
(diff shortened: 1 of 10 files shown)


helpers/strReplace.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const factory = () => {
+    return function(str, substr, newSubstr, iteration) {
+        if (typeof str !== 'string' || typeof substr !== 'string' || typeof newSubstr !== 'string') {
```
jairo-bc, line 5: In case you want to use this helper with concat or replace helper this condition won't pass. 
To unlock that possibility, please use https://github.com/bigcommerce/paper-handlebars/blob/master/helpers/lib/common.js#L12 before type checking.

helpers/strReplace.js
```diff
@@ -0,0 +1,43 @@
...
+            return 'Invalid Input';
```
jairo-bc, line 6: I think you would rather want to see here an Error, rather than a "Invalid Input" line. 
Please take a look at TypeError and maybe passing some arguments there, so it will be easier to debug

jairo-bc (changes requested):
Added few suggestions, but otherwise the approach looks good to me

thearcticalex:
Updated the above comments and created a new MR [here](https://github.com/bigcommerce/paper-handlebars/pull/157).

spec/helpers/strReplace.js
```diff
@@ -0,0 +1,67 @@
...
+                output: 'My name is Alex Alex Alex',
+            },
+        ], done);
+    });
+
+    it('should only handle string', function(done) {
+         runTestCases([
+            {
+                input: '{{strReplace 5 5 5}}',
+                output: 'Invalid Input',
```
jairo-bc, line 59: I think this test case also should be reworked to accommodate errors throwing

thearcticalex: Yes sir! Updated test case in the new commit.

jairo-bc (changes requested):

spec/helpers/strReplace.js
```diff
@@ -0,0 +1,68 @@
...
+            },
+            {
+                input: '{{strReplace string substr newSubstr 100}}',
+                output: 'My name is Alex Alex Alex',
+            },
+        ], done);
+    });
+
+    it('should only handle string', function(done) {
+        renderString('{{strReplace object "none" "Bob"}}').catch(e => {
```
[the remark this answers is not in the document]
thearcticalex: @jairo-bc Hi! I have updated the wording for the description. For catching errors, I am using the same format as other helpers, can you give me more details about how I can improve this please? Thank you!

PR #158, opened 2022-02-17
STRF-9658 Add log level variable to control log helper

jairo-bc:
## What? Why?

To get ability to log errors from paper and handlebars and fix the issue that caused https://jira.bigcommerce.com/browse/STRF-7500

## How was it tested?

npm test, e2e

----

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,8 +35,9 @@ class HandlebarsRenderer {
     * @param {Object} themeSettings - Theme settings (configuration), passed to helpers
     * @param {String} hbVersion - Which version of handlebars to use. One of ['v3', 'v4'] - defaults to 'v3'.
     * @param {Object} logger - A console-like object to use for logging
+    * @param {String} logLevel - log level which will be overriden by renderer
     */
-    constructor(siteSettings, themeSettings, hbVersion, logger = console) {
+    constructor(siteSettings, themeSettings, hbVersion, logger = console, logLevel = 'info') {
         // Figure out which version of Handlebars to use.
         switch(hbVersion) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -35,8 +35,9 @@ class HandlebarsRenderer {
     * @param {Object} themeSettings - Theme settings (configuration), passed to helpers
     * @param {String} hbVersion - Which version of handlebars to use. One of ['v3', 'v4'] - defaults to 'v3'.
     * @param {Object} logger - A console-like object to use for logging
+    * @param {String} logLevel - log level which will be overriden by renderer
     */
-    constructor(siteSettings, themeSettings, hbVersion, logger = console) {
+    constructor(siteSettings, themeSettings, hbVersion, logger = console, logLevel = 'info') {
```
jmwiese, line 40: Do we want to make this configurable env var? that would let us to adjust the level to debug in like integration if we wanted additional logging verbosity there.

jmwiese (approved):

index.js
```diff
@@ -338,6 +340,14 @@ class HandlebarsRenderer {
             }
         };
     }
+
+    /**
+     * 
+     * @param {String} level 
+     */
+    setLoggerLevel(level) {
+        this.handlebars.logger.level = level;
```
mattolson, line 349: Not sure that this will work for `console`. Can you write a test for that?

jairo-bc: sure, added! 
This behaviour is possible due to this if condition:
https://github.com/bigcommerce/paper-handlebars/blob/master/index.js#L332

Issue #111, opened 2020-11-03
incrementVar and decrementVar should not return a value


blacktlabs:
Either the documentation is wrong or incrementVar and decrementVar should not return the current value.
Returning the value is unexpected and results in unexpected text showing up on the web page they are used on.

pvaladez:
I just discovered this issue and filed a PR to remove the return lines in those helpers.  It's definitely undesirable to have the values of your incremented variables being printed out unexpectedly on the page.

As a workaround, you can use the `add` helper like this:
```handlebars
{{assignVar 'count' 0}}
{{assignVar 'count' (add (getVar 'count') 1)}}
```

jgorepp:
I've come across this and also agree that it would make more sense for these helpers to return a value

PR #160, opened 2022-02-18
Do not return values from incrementVar and decrementVar helpers

pvaladez:
## What? Why?
The issue was first reported in issue #111.

The incrementVar and decrementVar helpers currently return the value of the handlebars variable, which results in the value being outputted to html.

## How was it tested?
I removed the return lines on my local copy of stencil-cli/paper-handlebars, and the helpers then worked as expected without printing the variable values to html.

I'm not very familiar with the paper-handlebars source, but I can see that other helper functions like assignVar do not return values.

----

cc @bigcommerce/storefront-team @bookernath

### DIFF
--- a/helpers/decrementVar.js
+++ b/helpers/decrementVar.js
@@ -24,9 +24,6 @@ const factory = globals => {
             // Initialize or re-initialize value
             globals.storage.variables[key] = 0;
         }
-
-        // Return current value
-        return globals.storage.variables[key];
     };
 };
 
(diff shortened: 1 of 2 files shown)


jairo-bc:
@pvaladez I think some themes may rely on this kind of behaviour and this change might break backward compatibility.  
Maybe you can try to use assignVar or use incremental helpers in the exact moment, where the data should be displayed? 

cc @bookernath @mattolson Any thoughts?

PR #164, opened 2022-03-21
feat: STRF-9707 Improve money helper to support input params

jairo-bc:
## What? Why?

Improved money helper to support input params according to the docs

## How was it tested?

`npm test`, e2e

----

### DIFF
--- a/helpers/money.js
+++ b/helpers/money.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
-
 /**
  * Format numbers
  *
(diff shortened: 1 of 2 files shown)


spec/helpers/strReplace.js
```diff
@@ -56,6 +56,7 @@ describe('strReplace helper', function() {
 
     it('should throw an exception if the parameters have an invalid type', function(done) {
         renderString('{{strReplace object "none" "Bob"}}').catch(e => {
+            // console.log(e);
```
bc-max, line 59: do we need this comment here?

helpers/money.js
```diff
@@ -17,19 +15,32 @@ function numberFormat(value, n, s, c) {
...
 const factory = globals => {
-    return function(value) {
+    return function(...args) {
         const siteSettings = globals.getSiteSettings();
         const money = siteSettings.money;
 
-        if (!_.isNumber(value)) {
-            return '';
+        // remove options hash object
+        args.pop();
```
jmwiese, line 23: Are we able to exclude this by doing something like `return function(options, ...args)` to pull out the first param as "options" and just use the remaining in `args` to avoid this step?

jairo-bc: not really, cause I'm removing the last parameter, rather then the first one. 
So it should be like `return function(...args, options)`, but this kind of syntax is not supported 😒

jmwiese (approved):

helpers/money.js
jairo-bc: @jmwiese the thing is that number of arguments may vary. Just to be sure, that hash object (that has handlebars utility functions) was removed, I'm using a `pop()`. We are also using same approach in other helpers, to actually read that hash object by assigning it to a variable.

PR #171, opened 2022-06-15
fix: STRF-9873 use modified copies of `get`, `getObject`, `moment`, `option` 3p helpers

bc-evan-johnson:
## What? Why?

Bring a subset of 3rd-party helpers into our codebase with some modifications.

Internal details at https://bigcommercecloud.atlassian.net/browse/STRF-9873

## How was it tested?

Unit tests + interactive testing in CDVM.

----


helpers/get.js
```diff
@@ -0,0 +1,36 @@
...
+        if (arguments.length < 2) {
+            context = {};
+        }
+
+        // property name/path must be a string and a direct property of the object
+        if (typeof path !== 'string' || !_.has(context, path)) {
+            return;
+        }
+
+        let value = _.get(context, path);
```
jairo-bc, line 25: If it possible not to bring dependency on lodash, it would be ideal. We have a ticket in backlog to remove that dependency.

bc-evan-johnson: Done.

jairo-bc (approved):

jairo-bc (approved):
LGTM

PR #173, opened 2022-06-20
fix: STRF 9873 make `get` etc. compatible with `concat` return value

bc-evan-johnson:
## What? Why?

[#171](https://github.com/bigcommerce/paper-handlebars/pull/171) inadvertently broke an edge case where the `concat` helper is used to dynamically generate property paths.

With the original `get` and `option` helpers, this only worked with top-level properties (not nested properties) due to an implicit `toString` conversion [here](https://github.com/jonschlinkert/get-value/blob/2.0.6/index.js#L22). This PR modifies those helpers to accept both top-level and nested property paths generated by `concat`.

_Edit: During testing & review, I noticed that the old `get` helper would retrieve empty string keys (`get({'': 0}, '')` would return `0`) and have made changes to match that behavior._

## How was it tested?

Unit tests have been added to cover cases where `concat` is used to generate property paths for helpers modified in [#171](https://github.com/bigcommerce/paper-handlebars/pull/171).

----

### DIFF
--- a/helpers/lib/common.js
+++ b/helpers/lib/common.js
@@ -10,19 +10,25 @@ function isValidURL(val) {
 }
 
 /*
- * Based on https://github.com/jonschlinkert/get-value/blob/3.0.1/index.js, but
- * with configurability that was not used in handlebars-helpers removed.
+ * Based on https://github.com/jonschlinkert/get-value/blob/2.0.6/index.js with some enhancements.
+ * 
+ * - Performs "hasOwnProperty" checks for safety.
+ * - Now accepts Handlebars.SafeString paths.
  */
-function getValue(object, path) {
+function getValue(globals, object, path) {
     let parts;
+
+    // unwrap Handlebars.SafeString for compatibility with `concat` etc.
+    path = unwrapIfSafeString(globals.handlebars, path);
+
     // accept array or string for backwards compatibility
-    if (!Array.isArray(path)) {
-        if (typeof path !== 'string') {
-            return object;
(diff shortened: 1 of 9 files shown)


helpers/lib/common.js
```diff
@@ -10,21 +10,21 @@ function isValidURL(val) {
 }
 
 /*
- * Based on https://github.com/jonschlinkert/get-value/blob/3.0.1/index.js, but
```
jairo-bc, line 13: What was the intend to update this library? Looks like 2.0.6 is the current version in handlebars helpers.
Did you check if there are any breaking changes in behaviour?

helpers/lib/common.js
```diff
@@ -10,21 +10,21 @@ function isValidURL(val) {
...
+ * Based on https://github.com/jonschlinkert/get-value/blob/2.0.6/index.js, but
  * with configurability that was not used in handlebars-helpers removed.
  */
 function getValue(object, path) {
```
jairo-bc, line 16: Won't be it better to unwrap path here? What if someone uses this utility function with SafeString?

jairo-bc: Or maybe we can indicate that it doesn't support SafeString ( throw an error)

bc-evan-johnson: I've moved the unwrap call into `getValue` :)

jairo-bc (approved):
LGTM

jairo-bc (approved):

PR #178, opened 2022-07-05
bugfix & setup to replace 3p helpers incrementally

bc-evan-johnson:
## What? Why?

The previous iteration of these helper replacements broke backwards compatibility in a couple of previously-unknown edge cases.

Note: The helper replacements will be enabled in future PRs.

## How was it tested?

Additional unit test cases have been added. Both the new and old versions pass the new test cases.

----


spec/helpers/option.js
```diff
@@ -21,22 +21,23 @@ describe('option helper', function () {
...
 
-    it('accepts SafeString paths', (done) => {
-        runTestCases([
-            {
-                input: `{{option (concat 'a.b.' 'c')}}`,
-                output: `d`,
-            },
-        ], done);
-    });
+    // it('accepts SafeString paths', (done) => {
```
jairo-bc, line 35: Shouldn't it be supported now?

helpers.js
```diff
@@ -11,7 +11,7 @@ const helpersList = [
     'dynamicComponent',
     'encodeHtmlEntities',
     'for',
-    'get',
+    // 'get',
```
jairo-bc, line 14: Lets maybe enable 1 helper here? Maybe `get`

spec/helpers/option.js
bc-evan-johnson: The old `option` helper (and also `get`) only works with SafeString for top-level properties, not nested properties, so this test will fail on the old version.

helpers.js
bc-evan-johnson: Let's start with `moment` and `option`.

jairo-bc (approved):

PR #187, opened 2022-08-12
[STRF-9950] paper-handlebars - stylesheet helper early-hints flag

rafa-avila-bc:
JIRA: [STRF-9950](https://bigcommercecloud.atlassian.net/browse/STRF-9950)

## What? Why?

`stylesheet` helper is now able to produce resource hints.

## How was it tested?

uni tests

----

### DIFF
--- a/helpers/lib/resourceHints.js
+++ b/helpers/lib/resourceHints.js
@@ -1,12 +1,10 @@
-/* eslint-disable curly */
-const _ = require("lodash");
 const utils = require("handlebars-utils");
 
 const resourceHintsLimit = 50;
 
-const defaultResourceHintSate = 'preload';
+const defaultResourceHintState = 'preload';
 
-const resourceHintStates = [defaultResourceHintSate, 'preconnect', 'prefetch'];
+const resourceHintStates = [defaultResourceHintState, 'preconnect', 'prefetch'];
 
 const resourceHintFontType = 'font';
 const resourceHintStyleType = 'style';
@@ -20,12 +18,18 @@ const resourceHintAllowedTypes = [resourceHintStyleType, resourceHintFontType, r
  */
 function addResourceHint(globals, path, state, type) {
 
-    path = _.trim(path);
+    if (typeof path !== 'string') {
+        throw new Error('resourceHint attribute require a valid URI');
+    }
+    path = path.trim();
     if (path === '') {
-        return;
+        throw new Error('resourceHint received an empty path');
     }
 
-    state = _.trim(state);
+    if (typeof state !== 'string') {
+        throw new Error(`resourceHint attribute received invalid value. Valid values are: ${resourceHintStates}`);
+    }
+    state = state.trim();
     if (!resourceHintStates.includes(state)) {
         return;
     }
@@ -44,13 +48,14 @@ function addResourceHint(globals, path, state, type) {
         return;
     }
 
-    let value = Object.create({});
-    Object.defineProperty(value, 'src', {value: path, writable: false});
-    Object.defineProperty(value, 'state', {value: state, writable: false});
+    let value = {src: path, state};
 
-    type = _.trim(type);
-    if (type !== '' && _.includes(resourceHintAllowedTypes, type)) {
-        Object.defineProperty(value, 'type', {value: type, writable: false});
+    if (typeof type !== 'string') {
+        type = '';
+    }
+    type = type.trim();
+    if (type !== '' && resourceHintAllowedTypes.includes(type)) {
+        value.type = type;
     }
 
     globals.resourceHints.push(value);
@@ -58,7 +63,7 @@ function addResourceHint(globals, path, state, type) {
 
 module.exports = {
     resourceHintsLimit,
-    defaultResourceHintSate,
+    defaultResourceHintSate: defaultResourceHintState,
     addResourceHint,
     resourceHintAllowedTypes: {resourceHintStyleType, resourceHintFontType, resourceHintScriptType}
 };
--- a/helpers/lib/resourceHints.js
+++ b/helpers/lib/resourceHints.js
@@ -1,0 +1,64 @@
+/* eslint-disable curly */
+const _ = require("lodash");
+const utils = require("handlebars-utils");
+
+const resourceHintsLimit = 50;
+
+const defaultResourceHintSate = 'preload';
+
+const resourceHintStates = [defaultResourceHintSate, 'preconnect', 'prefetch'];
+
+const resourceHintFontType = 'font';
+const resourceHintStyleType = 'style';
+const resourceHintScriptType = 'script';
+const resourceHintAllowedTypes = [resourceHintStyleType, resourceHintFontType, resourceHintScriptType];
+
+/**
+ * @param {string} path - The uri to the resource.
+ * @param {string} state - 'preload' or 'preconnect'
+ * @param {string} type - any of [style, font, script]
+ */
+function addResourceHint(globals, path, state, type) {
+
+    path = _.trim(path);
+    if (path === '') {
+        return;
+    }
+
+    state = _.trim(state);
+    if (!resourceHintStates.includes(state)) {
+        return;
+    }
+
+    if (typeof globals.resourceHints === 'undefined') {
+        globals.resourceHints = [];
+    }
+
+    path = utils.escapeExpression(path);
+    let index = globals.resourceHints.findIndex(({src}) => path === src);
+    if (index >= 0) {
+        return;
+    }
+
+    if (globals.resourceHints.length >= resourceHintsLimit) {
+        return;
+    }
+
+    let value = Object.create({});
+    Object.defineProperty(value, 'src', {value: path, writable: false});
+    Object.defineProperty(value, 'state', {value: state, writable: false});
+
+    type = _.trim(type);
+    if (type !== '' && _.includes(resourceHintAllowedTypes, type)) {
+        Object.defineProperty(value, 'type', {value: type, writable: false});
+    }
+
+    globals.resourceHints.push(value);
+}
+
+module.exports = {
+    resourceHintsLimit,
+    defaultResourceHintSate,
+    addResourceHint,
+    resourceHintAllowedTypes: {resourceHintStyleType, resourceHintFontType, resourceHintScriptType}
+};
--- a/helpers/lib/resourceHints.js
+++ b/helpers/lib/resourceHints.js
@@ -18,20 +18,13 @@ const resourceHintAllowedTypes = [resourceHintStyleType, resourceHintFontType, r
  */
 function addResourceHint(globals, path, state, type) {
 
-    if (typeof path !== 'string') {
-        throw new Error('resourceHint attribute require a valid URI');
+    if (!utils.isString(path)) {
+        throw new Error('Invalid path provided. path should be a non empty string');
     }
     path = path.trim();
-    if (path === '') {
-        throw new Error('resourceHint received an empty path');
-    }
 
-    if (typeof state !== 'string') {
+    if (!utils.isString(state) || !resourceHintStates.includes(state)) {
         throw new Error(`resourceHint attribute received invalid value. Valid values are: ${resourceHintStates}`);
-    }
-    state = state.trim();
-    if (!resourceHintStates.includes(state)) {
-        return;
     }
 
     if (typeof globals.resourceHints === 'undefined') {
@@ -63,7 +56,7 @@ function addResourceHint(globals, path, state, type) {
 
 module.exports = {
     resourceHintsLimit,
-    defaultResourceHintSate: defaultResourceHintState,
+    defaultResourceHintState,
     addResourceHint,
     resourceHintAllowedTypes: {resourceHintStyleType, resourceHintFontType, resourceHintScriptType}
 };
--- a/helpers/stylesheet.js
+++ b/helpers/stylesheet.js
@@ -2,7 +2,7 @@
 
 const _ = require('lodash');
 const buildCDNHelper = require('./lib/cdnify');
-const ResourceHints = require('./lib/resourceHints');
+const {addResourceHint, resourceHintAllowedTypes} = require('./lib/resourceHints');
 
 const factory = globals => {
     return function (assetPath) {
@@ -20,11 +20,11 @@ const factory = globals => {
         const url = cdnify(path);
 
         if (_.has(options.hash, 'resourceHint')) {
-            ResourceHints.addResourceHint(
+            addResourceHint(
                 globals,
                 url,
                 options.hash['resourceHint'],
-                ResourceHints.resourceHintAllowedTypes.resourceHintStyleType
+                resourceHintAllowedTypes.resourceHintStyleType
             );
             delete options.hash.resourceHint;
         }
--- a/helpers/stylesheet.js
+++ b/helpers/stylesheet.js
@@ -2,9 +2,10 @@
 
 const _ = require('lodash');
 const buildCDNHelper = require('./lib/cdnify');
+const ResourceHints = require('./lib/resourceHints');
 
 const factory = globals => {
-    return function(assetPath) {
+    return function (assetPath) {
         const cdnify = buildCDNHelper(globals);
         const siteSettings = globals.getSiteSettings();
         const configId = siteSettings.theme_config_id;
@@ -18,12 +19,19 @@ const factory = globals => {
 
         const url = cdnify(path);
 
-        let attrs = { rel: 'stylesheet' };
+        if (_.has(options.hash, 'resourceHint')) {
+            ResourceHints.addResourceHint(
+                globals,
+                url,
+                options.hash['resourceHint'],
+                ResourceHints.resourceHintAllowedTypes.resourceHintStyleType
+            );
+            delete options.hash.resourceHint;
+        }
 
+        let attrs = {rel: 'stylesheet'};
         Object.assign(attrs, options.hash);
-
-        attrs = _.map(attrs, (value, key) => `${key}="${value}"`).join( ' ');
-
+        attrs = _.map(attrs, (value, key) => `${key}="${value}"`).join(' ');
         return `<link data-stencil-stylesheet href="${url}" ${attrs}>`;
     };
 };
--- a/helpers/stylesheet.js
+++ b/helpers/stylesheet.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const buildCDNHelper = require('./lib/cdnify');
 const {addResourceHint, resourceHintAllowedTypes} = require('./lib/resourceHints');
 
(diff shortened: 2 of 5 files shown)


Tiggerito:
How does this generate the resource hint link tags? Or is it part of supporting early hints via CloudFlare and HTTP header parameters (which would be cool)?

If it is early hints...

Adding the stylesheet is a great first move. And it looks like the code is designed to support other resources, like the fonts CSS.

Adding the fonts CSS and fonts script in early hints could mean the main fonts are loaded early. 

Supporting images would be of value. e.g. early preloading the page's main image. Would it be possible to add a helper to generate these early hints HTTP headers? Can they be placed anywhere in the theme? Being ably to add the early hint where the img is created would make things far more maintainable. e.g. where the product's main image is created, the early hint can be set. Or as an option in the responsive-img component.

Combine the above, and the resources required to render the page and reach LCP could be loaded before the HTML :-)

If it is not early hints...

Would it be beneficial to make it work with the existing helpers\resourceHints.js?

As the stylesheet is high priority and in the head, it starts to load instantly. This means any dns-prefetch or preconnect or preload also in the head probably has no benefit. 

These hints have value for resources that are later discovered or are after blocking resources. e.g. a preconnect to fonts.gstatic.com has value as it is not used until the fonts CSS file has loaded. But there is no value preconnecting to fonts.googleapis.com as the fonts CSS file is requested instantly.

Also, for compatibility with Safari, the dns-prefetch and preconnect should be separated and in a specific order. e.g.

&lt;link rel="preconnect" href="https://fonts.gstatic.com"  crossorigin="anonymous">
&lt;link rel="dns-prefetch" href="https://fonts.gstatic.com"  crossorigin="anonymous">

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable curly */
+const _ = require("lodash");
```
jairo-bc, line 2: Please, don't use `lodash`, we have a ticket in backlog to remove it from dependencies

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,64 @@
...
+const utils = require("handlebars-utils");
+
+const resourceHintsLimit = 50;
+
+const defaultResourceHintSate = 'preload';
```
jairo-bc, line 7:
```suggestion
const defaultResourceHintState = 'preload';
```

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,64 @@
...
+const resourceHintAllowedTypes = [resourceHintStyleType, resourceHintFontType, resourceHintScriptType];
+
+/**
+ * @param {string} path - The uri to the resource.
+ * @param {string} state - 'preload' or 'preconnect'
+ * @param {string} type - any of [style, font, script]
+ */
+function addResourceHint(globals, path, state, type) {
+
+    path = _.trim(path);
```
jairo-bc, line 23: `path.trim()`

jairo-bc, line 23: I would also recommend you to check if the `path` is a string, otherwise throw an error with some distinct message

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,64 @@
...
+        return;
+    }
+
+    if (globals.resourceHints.length >= resourceHintsLimit) {
+        return;
+    }
+
+    let value = Object.create({});
+    Object.defineProperty(value, 'src', {value: path, writable: false});
+    Object.defineProperty(value, 'state', {value: state, writable: false});
```
jairo-bc, line 49:
```suggestion
    const value = { src: path, state };
```

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,64 @@
...
+
+    type = _.trim(type);
+    if (type !== '' && _.includes(resourceHintAllowedTypes, type)) {
+        Object.defineProperty(value, 'type', {value: type, writable: false});
```
jairo-bc, line 53:
```suggestion
        value. type = type;
```

helpers/stylesheet.js
```diff
@@ -2,9 +2,10 @@
 
 const _ = require('lodash');
 const buildCDNHelper = require('./lib/cdnify');
+const ResourceHints = require('./lib/resourceHints');
```
jairo-bc, line 5: lowerCase please or you can directly import function by
`const { addResourceHint } = require('./lib/resourceHints');  `

spec/helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,140 @@
...
+        });
+
+        afterEach(done => {
+            sandbox.restore();
+            done();
+        });
+
+        it("does work as expected with valid params", (done) => {
+
+            let globals = {resourceHints: []};
```
jairo-bc, line 31: it's `const` and I think all above variables.
The main rule is if the variable is not reassigned (overwritten), so it's `const`.

spec/helpers/stylesheet.js
```diff
@@ -1,8 +1,11 @@
...
-      lab = exports.lab = Lab.script(),
-      describe = lab.experiment,
-      it = lab.it,
-      testRunner = require('../spec-helpers').testRunner;
+    lab = exports.lab = Lab.script(),
+    describe = lab.experiment,
+    it = lab.it;
+const {buildRenderer, testRunner} = require("../spec-helpers");
+const {expect} = require("code");
+const resourceHintStyleType = require("../../helpers/lib/resourceHints").resourceHintAllowedTypes.resourceHintStyleType;
```
jairo-bc, line 8:
```suggestion
const  { resourceHintStyleType }  = require("../../helpers/lib/resourceHints").resourceHintAllowedTypes;
```

spec/helpers/stylesheet.js
```diff
@@ -14,12 +17,27 @@ describe('stylesheet helper', () => {
...
-            {
-                input: '{{{stylesheet "assets/css/style.css"}}}',
-                output: '<link data-stencil-stylesheet href="https://cdn.bcapp/hash/stencil/123/css/style-xyz.css" rel="stylesheet">',
-            },
-        ], done);
+
+        const template = '{{{stylesheet "assets/css/style.css" resourceHint="preload"}}}';
+        const expected = '<link data-stencil-stylesheet href="https://cdn.bcapp/hash/stencil/123/css/style-xyz.css" rel="stylesheet">';
+
+        const renderer = buildRenderer(siteSettings);
```
jairo-bc, line 24: I think those test cases could be written in less code. 
1. testRunner on line 17 accepts `renderer` as a param. 
so you could wrap it with `beforeEach(() => {})`, so each time there would be a fresh copy of the `renderer`.
2. 2nd param of the `runTestCases()`, is a callback function, which is run at the end of the rendering (all promises). 
[Check this out](https://github.com/bigcommerce/paper-handlebars/blob/master/spec/spec-helpers.js#L47)
So you could add some your custom callbacks over there and after those at the end call `done();`

jairo-bc (changes requested):

rafa-avila-bc:
@Tiggerito the changes in this PR are not related to actual `link` tag generation but to "metadata" that will be used in early hints generation. So:

```
Supporting images would be of value. e.g. early preloading the page's main image.
```

- Would it be possible to add a helper to generate these early hints HTTP headers?
  Yes, the early hints work is in progress, and, yes, a new helper will exist for this purpose.
- Can they be placed anywhere in the theme?
  Yes. Helper won't depend on where in the theme it is used.

bookernath:
Just to add on to what @rafa-avila-bc mentioned, @Tiggerito, this is us evolving our approach to resource hints to allow you to register which resources should have preload HTTP headers, which is slightly faster (even by itself) than the hints which exist in HTML markup, for obvious reasons.

As you've already guessed, this further allows us to use HTTP Early Hints when clients have support for it, and Cloudflare helps with this.

We plan to add this ability to the `{{cdn}}` and `{{stylesheet}}` helpers to begin with, but we could certainly consider other resources as well. I want to test a bit to understand at what point preloading hits the point of diminishing returns or perhaps even causes worse performance; we do have to think about things like total HTTP header size for example, or how likely it is that resources will actually make meaningful progress on their downloading during the server response time. I expect there will need to be a balance here; preloading 100 things is probably unhelpful.

Our early assumption is that getting the theme JS bundle, stylesheet, and Script Manager scripts (of certain categories) preloaded is the right starting point.

Resources like images often benefit from e.g. progressive loading of JPGs so they _might_ benefit from preloading, or that precious time/bandwidth might be better spent somewhere else. We hope to find out through experimentation.

As a final note, a preload strategy doesn't play nicely with responsive images, because it's unclear at the time that HTTP headers are being generated what size image a client may need.

Tiggerito:
This is excellent news.

From my experience in improving Largest Contentful Paint (LCP) which is Google's measure related to speed and a user's page experience. The following can make a big difference:

Early loading of the main CSS, as it blocks rendering.
Early loading of the fonts CSS and theme-bundle.font.js, as it can trigger early loading of web fonts and reduce CLS issues.
Early loading of the images that can be LCP candidates (home page carousel, first products in a list, products main image, logo).

It does seem that early hints do not support responsive images. So they could not be included, which would typically mean the home page carousel image. But other images tend to be based on a fixed width, so the page always loads the same version of an image. 

Responsive images can typically still be preloaded. A mechanism to add preloads in the head from where the image is added would be cool. However, browsers should soon all be supporting importance="high" on the img tag, which should have the same effect and be far simpler to implement.

When those preloads are done, the main delays seem to be caused by the CPU being busy running scripts. I see improvements in LCP by actually delaying the main bundle and other CPU-intensive scripts so the CPU can get to rendering earlier. My tests have the script load early but run later.

Another improvement was related to the slider mechanism, which causes long-running scripts. I delayed it (no need to slide before the images are loaded) and split it into separate tasks per slider, so it did not stop user interaction.

I'm also seeing that the use of defer can be detrimental as it forces DCL to wait for a script to load and run. In contrast, async is not blocking like that. If a script does not need to run before the DOM is rendered, it should not be defer but async. 

A similar issue comes from scripts waiting on DCL. This can often be before rendering and thus delaying it. I'm experimenting with adding a timeout to slightly delay these scripts from running, so the CPU decides to render instead.

I find only a few scripts would benefit from preloading. Basically, ones that affect the look of the page. Like Google Optimize or widget scripts.

It may be of value to have it also add link tags in order to support other browsers.

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * @param {string} path - The uri to the resource.
+ * @param {string} state - 'preload' or 'preconnect'
+ * @param {string} type - any of [style, font, script]
+ */
+function addResourceHint(globals, path, state, type) {
+
+    if (typeof path !== 'string') {
+        throw new Error('resourceHint attribute require a valid URI');
```
jairo-bc, line 22: It's better to say that it should be a String, you are not checking here that it's a valid url

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,69 @@
...
+    if (path === '') {
+        throw new Error('resourceHint received an empty path');
+    }
+
+    if (typeof state !== 'string') {
+        throw new Error(`resourceHint attribute received invalid value. Valid values are: ${resourceHintStates}`);
+    }
+    state = state.trim();
+    if (!resourceHintStates.includes(state)) {
+        return;
```
jairo-bc, line 34: I think you can combine this check and above, cause this pretty much indicates that there was an `invalid`(not supported) value

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    if (typeof globals.resourceHints === 'undefined') {
+        globals.resourceHints = [];
```
jairo-bc, line 38: Maybe we create empty array list, when `globals` is initialized? I believe it happens in `index.js`

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,69 @@
...
+ * @param {string} path - The uri to the resource.
+ * @param {string} state - 'preload' or 'preconnect'
+ * @param {string} type - any of [style, font, script]
+ */
+function addResourceHint(globals, path, state, type) {
+
+    if (typeof path !== 'string') {
+        throw new Error('resourceHint attribute require a valid URI');
+    }
+    path = path.trim();
```
jairo-bc, line 24: Do we need to `trim` the values (here and below)? That's basically mean that we allow whitespaces in the values

spec/helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,151 @@
...
+            sandbox = Sinon.createSandbox();
+            done();
+        });
+
+        afterEach(done => {
+            sandbox.restore();
+            done();
+        });
+
+        it("does work as expected with valid params", (done) => {
```
jairo-bc, line 33: Could you please name the test as what it does? It's not really clear from the test name what is it doing or what is it behaviour

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,69 @@
...
+    if (type !== '' && resourceHintAllowedTypes.includes(type)) {
+        value.type = type;
+    }
+
+    globals.resourceHints.push(value);
+}
+
+module.exports = {
+    resourceHintsLimit,
+    defaultResourceHintSate: defaultResourceHintState,
```
jairo-bc, line 66:
```suggestion
    defaultResourceHintSate,
```

helpers/lib/resourceHints.js
jairo-bc: Also, do you think it makes sense to rename this variable to `resourceHintsHeaders`, since it's used only headers?

helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    if (globals.resourceHints.length >= resourceHintsLimit) {
+        return;
+    }
+
+    let value = {src: path, state};
+
+    if (typeof type !== 'string') {
+        type = '';
```
jairo-bc, line 54: How should it work? 
I think this is the use case `{{{stylesheet "bla-bla.scss" resourceHint=""}}}'`, which will be converted to 
`"Link: <https://cdn.com/bla-bla.scss>;"`. Will be there an effect without `rel` attribute?

helpers/stylesheet.js
```diff
@@ -18,12 +19,19 @@ const factory = globals => {
 
         const url = cdnify(path);
 
-        let attrs = { rel: 'stylesheet' };
-
+        if (_.has(options.hash, 'resourceHint')) {
```
jairo-bc, line 22: lodash 🥲
you can try to do 
```suggestion
        if (options.hash.resourceHint) {
```

helpers/stylesheet.js
```diff
@@ -18,12 +19,19 @@ const factory = globals => {
...
+                globals,
+                url,
+                options.hash['resourceHint'],
+                resourceHintAllowedTypes.resourceHintStyleType
+            );
+            delete options.hash.resourceHint;
+        }
+
+        let attrs = {rel: 'stylesheet'};
         Object.assign(attrs, options.hash);
```
jairo-bc, line 33:
```suggestion
        const attrs = {rel: 'stylesheet', ...options.hash};
```

helpers/stylesheet.js
```diff
@@ -18,12 +19,19 @@ const factory = globals => {
...
-
-        attrs = _.map(attrs, (value, key) => `${key}="${value}"`).join( ' ');
-
+        attrs = _.map(attrs, (value, key) => `${key}="${value}"`).join(' ');
```
jairo-bc, line 34: you can try to use `for.. in` here, and I also don't recommend to override `attrs` variable

spec/helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,151 @@
...
+            const type = 'style';
+            const state = 'preload';
+            const expected = {src, state, type};
+
+            addResourceHint(globals, src, state, type);
+
+            expect(globals.resourceHints.push.called).to.equals(true);
+            expect(globals.resourceHints).to.have.length(1);
+            const result = _.head(globals.resourceHints);
+            expect(result).to.equals(expected);
```
jairo-bc, line 48: I think this will be sufficient
```suggestion
            expect(globals.resourceHints[0]).to.equals(expected);
```

jairo-bc: and you won't need anymore to spy on `globals.resourceHints.push`

spec/helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,151 @@
...
+                    );
+                }
+                return f;
+            });
+
+            throws.forEach(t => {
+                expect(t).to.throw();
+            });
+
+            expect(globals.resourceHints.push.notCalled).to.equals(true);
```
jairo-bc, line 73:
```suggestion
            expect(globals.resourceHints.length).to.equals(0);
```
and you won't need to spy on `push` happened

spec/helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,151 @@
...
+                        defaultResourceHintSate,
+                        resourceHintAllowedTypes.resourceHintStyleType
+                    );
+                }
+                return f;
+            });
+
+            throws.forEach(t => {
+                expect(t).to.throw();
+            });
```
jairo-bc, line 71: I think it can work like this
```suggestion
            [undefined, null, ''].forEach(input => {
                expect(
                    addResourceHint(
                        global,
                        input,
                        defaultResourceHintSate,
                        resourceHintAllowedTypes.resourceHintStyleType
                    );
                ).to.throw();
            });
```

spec/helpers/lib/resourceHints.js
```diff
@@ -0,0 +1,151 @@
...
+
+            addResourceHint(
+                globals,
+                'https://my.asset.css',
+                defaultResourceHintSate
+            );
+
+            expect(globals.resourceHints.push.calledOnce).to.equals(true);
+            const hint = _.head(globals.resourceHints);
+            expect(hint.type).to.not.exist();
```
jairo-bc, line 90: I think you can change this test case (and those below) to not use `lodash` and `push` spy

spec/helpers/stylesheet.js
```diff
@@ -14,12 +16,23 @@ describe('stylesheet helper', () => {
     const runTestCases = testRunner({siteSettings});
 
     it('should render a link tag with the cdn ulr and stencil-stylesheet data tag', done => {
-        runTestCases([
+        const renderer = buildRenderer(siteSettings);
+        const runner = testRunner({renderer});
```
jairo-bc, line 20: I think it's possible to do something like this (before all test cases)
```
describe('....', () => {
  let renderer, runner;
  beforeEach(() => {
     renderer = buildRenderer(siteSettings);
     runner = testRunner({renderer});
  });
  
  it('test 1', () => {
     runner([], done);
  });
   it('test 2', () => {
     runner([], done);
  });
}
```
So before each test case new instance of `renderer` and `runner` will be created and the state won't be preserved

jairo-bc (changes requested):

jairo-bc (approved):
LGTM

PR #188, opened 2022-08-18
[STRF-9948] - paper-handlebars - resourceHints helper add new early-hints flag param

rafa-avila-bc:
## What? Why?

1. By default data for resource hints is generated for anything the `resourceHints` helper generates `link` tags, right now only font providers.
2. `getFontsCollection` helper takes in an optional `resourceHint='{REL}'` attribute indicating whether or not to generate data resource hints.  

## How was it tested?

unit tests

----

### DIFF
--- a/helpers/lib/resourceHints.js
+++ b/helpers/lib/resourceHints.js
@@ -1,12 +1,10 @@
-/* eslint-disable curly */
-const _ = require("lodash");
 const utils = require("handlebars-utils");
 
 const resourceHintsLimit = 50;
 
-const defaultResourceHintSate = 'preload';
+const defaultResourceHintState = 'preload';
 
-const resourceHintStates = [defaultResourceHintSate, 'preconnect', 'prefetch'];
+const resourceHintStates = [defaultResourceHintState, 'preconnect', 'prefetch'];
 
 const resourceHintFontType = 'font';
 const resourceHintStyleType = 'style';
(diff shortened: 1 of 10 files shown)


helpers/lib/resourceHints.js
```diff
@@ -41,22 +48,30 @@ function addResourceHint(globals, path, state, type) {
...
-    if (typeof type !== 'string') {
-        type = '';
+    if (utils.isString(type) && resourceHintTypes.includes(type)) {
+        hint.type = type;
     }
-    type = type.trim();
-    if (type !== '' && resourceHintAllowedTypes.includes(type)) {
-        value.type = type;
+
+    if (!utils.isString(cors) || !allowedCors.includes(cors)) {
```
jairo-bc, line 57: don't you think it's better to throw an error if we use some random cors value, rather then fallback to "no"?

helpers/resourceHints.js
```diff
@@ -10,27 +16,38 @@ const fontResources = {
     ],
 };
 
+function format(host) {
```
jairo-bc, line 19: Nice optimization :)

helpers/resourceHints.js
```diff
@@ -10,27 +16,38 @@ const fontResources = {
...
+
 const factory = globals => {
-    return function() {
-        function format(host) {
-            return `<link rel="dns-prefetch preconnect" href="${host}" crossorigin>`;
-        }
+    return function () {
 
-        var hosts = [];
+        let hosts = [];
```
jairo-bc, line 26: it's `const`

helpers/lib/resourceHints.js
rafa-avila-bc: you are right. we should default to cors only when no value is provided as that is the default a link tag and link headers have.

helpers/resourceHints.js
rafa-avila-bc: maybe it is not because we are reassigning it at line [39](https://github.com/bigcommerce/paper-handlebars/pull/188/files#diff-b2e5ac345eb2c02e97a25341cda5e936b4952130cfa11ba6e10e44d9479ca03dR39)

jairo-bc (approved):

PR #197, opened 2022-09-20
[STRF-10096] - storefront - investigate google fonts css uri encoding issue

rafa-avila-bc:
## What? Why?

After fully enabling early hints an issue with URL encoding was found: the URL for google fonts in Link headers are encoded but the same URLs in the `link` tag are not encoded, leading to requesting the fonts 2 times. Check the requests in this test https://www.webpagetest.org/result/220912_BiDcW5_D8J/1/details/#waterfall_view_step1

The proposed fix in this PR is that the code responsible for creating the data for early hints will use NodeJs `url` to `parse` the paths and the `format`ed version of the valid URL will be returned so any helper using this code will then use the valid URL to generate the HTML it needs to. The `url.parse` method throws when it receives invalid data so it already helps us to produce valid URLs. It is important to note that NodeJs `url` module does not follow RFC-3986 URI model. If there is a need to comply with RFC-3986, a new package should be installed: check: https://runkit.com/embed/hrxz28cj47n1 

## How was it tested?

Unit tests.

----

### DIFF
--- a/helpers/lib/fonts.js
+++ b/helpers/lib/fonts.js
@@ -1,8 +1,9 @@
 'use strict';
 
 const _ = require('lodash');
+const URL = require('url')
 
-const {resourceHintAllowedTypes, addResourceHint} = require('../lib/resourceHints');
+const { resourceHintAllowedTypes, addResourceHint } = require('../lib/resourceHints');
 
 const fontProviders = {
     'Google': {
@@ -19,7 +20,7 @@ const fontProviders = {
          *
          * @returns {string}
          */
-        parser: function(fonts) {
+        parser: function (fonts) {
             var collection = [],
                 familyHash = {};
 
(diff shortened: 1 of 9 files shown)


helpers/lib/fonts.js
```diff
@@ -53,13 +54,19 @@ const fontProviders = {
...
+        buildLink: function (fonts, fontDisplay) {
             const displayTypes = ['auto', 'block', 'swap', 'fallback', 'optional'];
             fontDisplay = displayTypes.includes(fontDisplay) ? fontDisplay : 'swap';
-            return `<link href="https://fonts.googleapis.com/css?family=${fonts.join('|')}&display=${fontDisplay}" rel="stylesheet">`;
+            const uri = `https://fonts.googleapis.com/css?family=${fonts.join('|')}&display=${fontDisplay}`
+            try {
+                const url = URL.parse(uri);
+                return `<link href="${url.format()}" rel="stylesheet">`;
+            } catch (error) {
+                throw new Error(`Invalid URL [${uri}]. Check configured fonts.`)
```
jairo-bc, line 65: Do you have examples of invalid urls that occurred? Maybe we can cover this case with unit tests

helpers/lib/resourceHints.js
```diff
@@ -30,6 +31,11 @@ function addResourceHint(globals, path, state, type, cors) {
         throw new Error('Invalid path provided. path should be a non empty string');
     }
     path = path.trim();
+    try {
+        path = URL.parse(path).format();
+    } catch (error) {
+        throw new Error(`Invalid value (resource URL) provided: [${path}]`)
```
jairo-bc, line 37: You may want to use you may want just to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent

If you really need to validate the url. There is https://github.com/bigcommerce/paper-handlebars/blob/master/helpers/lib/common.js#L4

helpers/lib/fonts.js
rafa-avila-bc: yeah. I am not aware of any case of invalid URLs yet. Maybe because we had no way to know that given that only the browser was actually doing something with those URLs. Will add coverage to this code path.

helpers/lib/resourceHints.js
rafa-avila-bc: Using the `common` utilities, in this case, is not an option given it does not take into account valid relative URI.

```js
❯ node
Welcome to Node.js v14.20.0.
Type ".help" for more information.
> const common = require('./helpers/lib/common')
undefined
> common.isValidURL('/partial.js?tm=1234567890')
false
>
```

`decodeURIComponent` is not an option either because if `URL.parse` when given an already encoded URL does not double encode it.

```js
> URL.parse(decodeURIComponent('https://fonts.googleapis.com/css?family=Karla:400|Montserrat:400,500,700&display=block')).format()
'https://fonts.googleapis.com/css?family=Karla:400%7CMontserrat:400,500,700&display=block'
```

jairo-bc (approved):

jairo-bc (dismissed):

jmwiese (approved):

PR #198, opened 2022-09-21
Add original stack trace to render error

jmwiese:
## What? Why?

We're seeing a lot of errors about 'hash' undefined in renderer that spike occasionally and result in 500s https://kibana.bigcommerce.net/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2d,to:now))&_a=(columns:!(template_path,template_engine,message,store_id),filters:!(),index:production_containers_storefront-renderer,interval:auto,query:(language:kuery,query:'%22Cannot%20read%20property%20!'hash!'%20of%20null%22'),sort:!(!('@timestamp',desc)))

We're attempting to diagnose what exactly is failing within handlebars when we have a RenderError, so add the original stack trace to the error message 

## How was it tested?

----

### DIFF
--- a/index.js
+++ b/index.js
@@ -277,7 +277,7 @@ class HandlebarsRenderer {
             try {
                 result = template(context);
             } catch(e) {
-                return reject(new RenderError(e.message));
+                return reject(new RenderError(`${e.message} : ${e.stack}`));
             }
 
             // Apply decorators


rafa-avila-bc (approved):

jairo-bc:
Where you able to get it locally ? @jmwiese 
I think [AppError](https://github.com/jmwiese/paper-handlebars/blob/eba2dd15a0180c5f70d5adba955052f87523111b/lib/appError.js#L12) should be doing it, but for some reason it doesn't work. 
Maybe you can remove the `this.construtor` over there?

jmwiese:
@jairo-bc For the stack trace in kibana we're only seeing the stack trace for throwing of AppError

```
RenderError: Cannot read property 'hash' of null     at /opt/app/node_modules/@bigcommerce/stencil-paper-handlebars/index.js:280:31     at new Promise (<anonymous>)     at HandlebarsRenderer.render (/opt/app/node_modules/@bigcommerce/stencil-paper-handlebars/index.js:260:16)     at Paper.render (/opt/app/node_modules/@bigcommerce/stencil-paper/index.js:226:30)     at /opt/app/server/services/multiPhaseRenderer.js:94:28     at runMicrotasks (<anonymous>)     at processTicksAndRejections (internal/process/task_queues.js:95:5)
--



```
 it doesnt seem to include the root cause of the exception from `template(context)` handlebars call

jairo-bc (dismissed):

PR #199, opened 2022-09-26
STRF-10061 remove handlebars-utils dependency

zackangelo:
## What? Why?

* Finishes the removal of `handlebars-utils`
* Added missing functions (e.g, `isObject`) to `utils/index.js`
* Removed circular dependencies from `markdown` and `createFrame` modules

Also, renamed `utils.js` to `base.js` because it was confusing to have a submodule the same name as its parent module.

## How was it tested?

`npm run test`

----

### DIFF
--- a/helpers/3p/utils/lib/createFrame.js
+++ b/helpers/3p/utils/lib/createFrame.js
@@ -4,9 +4,17 @@
  * https://github.com/jonschlinkert/define-property/blob/master/index.js
  */
 
+function extend(obj /* , ...source */) {
+  for (var i = 1; i < arguments.length; i++) {
+    for (var key in arguments[i]) {
+      if (Object.prototype.hasOwnProperty.call(arguments[i], key)) {
+        obj[key] = arguments[i][key];
+      }
+    }
+  }
 
-const utils = require('handlebars-utils');
-
+  return obj;
+}
 
 function defineProperty(obj, prop, val) {
   if (typeof obj !== 'object' && typeof obj !== 'function') {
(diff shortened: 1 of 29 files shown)


helpers/3p/utils/lib/createFrame.js
```diff
@@ -4,9 +4,17 @@
  * https://github.com/jonschlinkert/define-property/blob/master/index.js
  */
 
+function extend(obj /* , ...source */) {
```
zackangelo, line 7: not sure why this fn was in `utils`, the only place it appears to be used is `createFrame`. So moved it here to eliminate the circular dependency.

helpers/3p/utils/lib/createFrame.js
```diff
@@ -27,15 +35,14 @@ function defineProperty(obj, prop, val) {
 };
 
 module.exports = function createFrame(data) {
-  if (!utils.isObject(data)) {
+  if (typeof data !== 'object') {
```
zackangelo, line 38: Use `typeof` instead to eliminate circular dependency.

helpers/3p/utils/lib/markdown.js
```diff
@@ -22,7 +21,7 @@ module.exports = function markdown(config) {
       return md.render(context);
     }
 
-    if (utils.isObject(context) && typeof context.fn === 'function') {
+    if (typeof context === 'object' && typeof context.fn === 'function') {
```
zackangelo, line 24: Use `typeof` instead to eliminate circular dependency.

zackangelo:
Superceded by #209

PR #201, opened 2022-10-06
fix: STRF-10122 Fix inArray helper, when array is undefined

jairo-bc:
## What? Why?

Original error:
 Cannot read property 'length' of undefined

Happened in inArray helper, when array is undefined.

## How was it tested?

npm test

----


jordanarldt (approved):
Looks good 👍

PR #212, opened 2022-10-18
[STRF-10120] paper-handlebars - resource hints should include the proper values for PreloadResourceType

rafa-avila-bc:
## What? Why?

- fixes issues of duplicated hint returning undefined.

## How was it tested?

- unit tests

----


helpers/lib/resourceHints.js
```diff
@@ -48,11 +48,11 @@ function addResourceHint(globals, path, rel, type, cors) {
 
     let index = globals.resourceHints.findIndex(({src}) => path === src);
     if (index >= 0) {
-        return;
+        return path;
```
rafa-avila-bc, line 51: There may be scenarios where nested templates include helpers that produce preload hints for the same resource where only the 1st usage of the helper will have the correct effect and the rest, will have unexpected results.

As an example let's take `stylesheet`, also, think that there is 1 `common-resources` (nested) template that is included in a `base` template. The nested template includes the `{{{ stylesheet '/asset/css/theme.css' }}}` helper which `base` template also includes but, in a linear order, the nested template is rendered before the occurrence of the `stylesheet` helper in the base template. For the nested template, the produced `link` tag will be correct, it will include a valid `href` attribute. The `stylesheet` in the `base` template:
- (with `return;`) will produce a `href="undefined"`
- (with the proposed change) will produce a valid `href` attribute, with the correct URL.

spec/helpers/stylesheet.js
```diff
@@ -37,6 +37,35 @@ describe('stylesheet helper', () => {
         });
     });
 
+    it('should render both link tags correclty but produce only one resource hint', done => {
```
rafa-avila-bc, line 40: This test exemplifies what's described in this comment https://github.com/bigcommerce/paper-handlebars/pull/212/files#r998713714

helpers/lib/resourceHints.js
```diff
@@ -48,11 +48,11 @@ function addResourceHint(globals, path, rel, type, cors) {
...
     }
 
     if (globals.resourceHints.length >= resourceHintsLimit) {
-        return;
+        throw new Error(`Resource Hints limit (${resourceHintsLimit}) reached.`);
```
jairo-bc, line 55: Are we good to throw the error even it can be not under full user control in cases when cdn, stylesheet helpers are being used?

rafa-avila-bc: `cdn` is not adding reload hint by default but `stylesheet` and `getFontsCollection`. But yeah, you are right, given that we are exposing adding hints as default for a couple of helpers, we better no to throw here but do nothing.

jairo-bc (approved):

jairo-bc (approved):

PR #232, opened 2022-12-01
fix: STRF-10188 adds logs at early hint generation calls

rafa-avila-bc:
## What? Why?

Jira: https://bigcommercecloud.atlassian.net/browse/STRF-10188

adds logs for errors at early-hint generation calls.

## How was it tested?

given that only logs were added, unit tests are our safety net here.

----

### DIFF
--- a/helpers/cdn.js
+++ b/helpers/cdn.js
@@ -23,7 +23,7 @@ const factory = globals => {
         const options = arguments[arguments.length - 1];
         if (options.hash && options.hash.resourceHint) {
             if (!utils.isString(fullPath)) {
-                console.warn(`Skipping resource hint creation in CDN fill path was not valid [${fullPath}]`);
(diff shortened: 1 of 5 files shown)


helpers/cdn.js
```diff
@@ -2,21 +2,37 @@
...
         const options = arguments[arguments.length - 1];
         if (options.hash && options.hash.resourceHint) {
-            fullPath = addResourceHint(
-                globals,
-                fullPath,
-                options.hash.resourceHint,
-                options.hash.as
-            );
+            if (!utils.isString(fullPath)) {
+                console.warn(`Skipping resource hint creation in CDN fill path was not valid [${fullPath}]`);
```
jmwiese, line 26: I think if these are logging out to console as WARN in renderer it will trigger a sentry alert, do we want this in this case?

rafa-avila-bc: only in the `earlyHint` helper case. will downgrade all `warn` no `info` tho.

PR #235, opened 2022-12-21
feat: Add JSONparseSafe helper

RomanKrasinskyi:
## What? Why?
We have the JSONParse helper which parse JSON and return an object:
```
{{#JSONparse jsonString}}
  {{name}}
{{/JSONparse}}
```
Helper works fine until non-json comes to parse. In this case it will break a page because it always expects syntactically correct JSON.

We faced with such case when parameter can be string or json and need to have a helper that can parse it safely.
According to required, we want to add a new `JSONparseSafe` helper:
```
{{#JSONparseSafe jsonString}}
  {{name}}
{{/JSONparseSafe}}

{{#JSONparseSafe jsonString}}
  {{name}}
{{else}}
  {{string}}
{{/JSONparseSafe}}
```
So, the helper try to parse passed parameter then
  - if succeeded - return an object
  - if failure - do nothing
  - if failure and we have else instruction - do else instruction
  
If you need more information please reach me

## How was it tested?
Tested manually / Unit tests

----

### DIFF
--- a/spec/helpers/jsonParseSafe.js
+++ b/spec/helpers/jsonParseSafe.js
@@ -1,0 +1,28 @@
+const { describe, it } = exports.lab = require('lab').script();
+
+const { testRunner } = require('../spec-helpers');
+
+describe('JSONParseSafe helper', () => {
+    const context = {
+        jsonString: '{"name": "John"}',
+        string: 'John Doe',
+    };
+    const runTestCases = testRunner({context});
+
+    it('should have the same behavior as the original if helper', done => {
+        runTestCases([
+            {
+                input: '{{#JSONparseSafe string}}{{name}}{{/JSONparseSafe}}',
+                output: '',
+            },
+            {
+                input: '{{#JSONparseSafe jsonString}}{{name}}{{/JSONparseSafe}}',
+                output: 'John',
+            },
+            {
+                input: '{{#JSONparseSafe string}}{{name}}{{else}}{{string}}{{/JSONparseSafe}}',
+                output: 'John Doe',
+            },
+        ], done);
+    });
+});
--- a/spec/helpers/jsonParseSafe.js
+++ b/spec/helpers/jsonParseSafe.js
@@ -9,16 +9,26 @@ describe('JSONParseSafe helper', () => {
     };
     const runTestCases = testRunner({context});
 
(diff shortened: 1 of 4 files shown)


BC-krasnoshapka (approved):

spec/helpers/jsonParseSafe.js
```diff
@@ -0,0 +1,28 @@
...
+    const context = {
+        jsonString: '{"name": "John"}',
+        string: 'John Doe',
+    };
+    const runTestCases = testRunner({context});
+
+    it('should have the same behavior as the original if helper', done => {
+        runTestCases([
+            {
+                input: '{{#JSONparseSafe string}}{{name}}{{/JSONparseSafe}}',
```
jairo-bc, line 15: I think this case is outside of the original helper

jairo-bc:
cc @bigcommerce/dev-docs

spec/helpers/jsonParseSafe.js
RomanKrasinskyi: Updated tests

BC-krasnoshapka (approved):

bookernath:
Seems fine to me - @jairo-bc WDYT?

@RomanKrasinskyi it occurs to me from your description that perhaps the existing `{{JSONparse}}` helper should have a better failure mode, such as silently returning an empty string if the JSON fails to parse or only logging a warning. It's not great to have helpers that can completely break a page with bad input.

If we had such a thing, you could also do something like:

```
{{#if (JSONparse something) }}
  {{JSONparse something}}
{{#else}}
  <!-- do something else -->
{{/if}}
```

I have no problem with also creating this new helper, I just think we should fix the safety on the old one also. :)

bookernath (approved):

bc-traciporter:
Created a ticket to implement these changes. See [DEVDOCS-4750](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4750).

RomanKrasinskyi:
@bookernath
It's a good point to change behaviour of the existing helper. I thought to change {{JSONparse}} behaviour like in a new one, but we can break some thing. So we decided to create a new one.

In your suggestion we will parse json twice: once in condition and another one inside {{#if}} block. It look strange.

jairo-bc (approved):

PR #245, opened 2023-02-08
fix: STRF-9673 Lodash removal, part 3

jairo-bc:
## What? Why?
Helpers: pluck, after, truncate

## How was it tested?

npm test

----

### DIFF
--- a/helpers/truncate.js
+++ b/helpers/truncate.js
@@ -25,7 +25,7 @@ const substring = require('stringz').substring;
  */
 const factory = globals => {
     return function(string, length) {
-        if (typeof string !== 'string') {
+        if (typeof string !== 'string' || string.length === 0) {
             return string;
         }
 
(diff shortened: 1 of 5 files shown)


helpers/truncate.js
```diff
@@ -25,7 +25,7 @@ const substring = require('stringz').substring;
  */
 const factory = globals => {
     return function(string, length) {
-        if (typeof string !== 'string') {
+        if (typeof string !== 'string' || string.length === 0) {
```
bc-max, line 28: So if the `string` is an array or object (is it possible case?) we just return it. What if it such an array/object is too big?

jairo-bc: that's right, that's part of user input validation. The helper supports only non-empty strings. 
What do you mean by too big array/object? The data is either returned by backend or determined by user (meaning in templates). We have limitation on bundle size ( and I think on a template size)

rafa-avila-bc (approved):

PR #252, opened 2023-03-15
fix: STRF-10494 last helper works with strings

rafa-avila-bc:
## What? Why?

`last` helper now supports strings as the first argument.

## How was it tested?

- unit tests

----


bc-max (dismissed):

jairo-bc:
@rafa-avila-bc I think they also mentioned about `first` helper in the ticket. Do you mind also check if covering those scenarios in public docs?

bc-max (approved):

PR #256, opened 2023-03-16
Allow strings to be replaced with an empty string. Changes within strReplace file

partner91:
## What? Why?

Added check for an empty string to be eligible to change the old string.  This feature would allow you to easily manipulate prices within BC templates because you have to strip currency sign commas etc.

## How was it tested?

This feature was tested on the development BigCommerce theme and it works as expected. 

----

### DIFF
--- a/helpers/strReplace.js
+++ b/helpers/strReplace.js
@@ -14,7 +14,7 @@ const factory = globals => {
             throw new ValidationError("Invalid query parameter string passed to strReplace");
         } else if (!utils.isString(substr)) {
             throw new ValidationError("Invalid query paramter substring passed to strReplace");
-        } else if (!utils.isString(newSubstr)) {
+        } else if (!utils.isString(newSubstr) && newSubstr !== "") {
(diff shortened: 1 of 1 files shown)


jairo-bc:
@partner91 do you mind providing unit tests for your improvement that covers before and after cases?

partner91:
In spec/helpers/strReplace.js there is a test for an empty string, somehow it runs without throwing an error. But without adding check that I've posted you can't replace string with an empty character.  You can try running all tests.  Below you can find the test I'm referring to. 

 it('should remove characters from string', function(done) {
        runTestCases([
            {
                input: '{{strReplace "123-45-6789" "-" ""}}',
                output: '123456789',
            },
        ], done);
    });

jairo-bc:
Hi @partner91, this feature should be addressed by https://github.com/bigcommerce/paper-handlebars/pull/259. Could you please verify at your end?

jairo-bc:
Closing this since it was fixed in https://github.com/bigcommerce/paper-handlebars/pull/259

PR #266, opened 2023-04-06
fix: STRF-10493 Address markdown helper messy logs

rafa-avila-bc:
## What? Why?

JIRA: [STRF-10493](https://bigcommercecloud.atlassian.net/browse/STRF-10493)

`markdown` helper produces multi-line logs that are not even needed in debug mode so we remove those logs and enable the `Remarkable` lib configuration.

## How was it tested?

Unit tests for `Remarkable` configuration were enabled.
All unit tests are green.

----

[STRF-10493]: https://bigcommercecloud.atlassian.net/browse/STRF-10493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helpers/3p/utils/lib/markdown.js
+++ b/helpers/3p/utils/lib/markdown.js
@@ -1,41 +1,49 @@
 const merge = require('./mixinDeep');
-const { Remarkable } = require('remarkable');
+const {Remarkable} = require('remarkable');
+const {linkify} = require('remarkable/linkify');
 
 module.exports = function markdown(config) {
 
-  if (typeof config === 'string') {
-    return helper.apply(this, arguments);
-  }
-
-  config = config || {};
-  if (config.fn || config.hash || arguments.length > 1) {
-    console.log(arguments);
-    return helper.apply(this, arguments);
-  }
-
-  function helper(context, options) {
-    if (typeof context === 'string') {
-      var opts = merge({}, config, options);
-      console.log(options)
-      var md = new Remarkable(opts);
-      return md.render(context);
+    if (typeof config === 'string') {
+        return helper.apply(this, arguments);
     }
 
-    if (typeof context === 'object' && typeof context.fn === 'function') {
-      options = context;
-      context = {};
+    config = config || {};
+    if (config.fn || config.hash || arguments.length > 1) {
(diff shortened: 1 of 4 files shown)


spec/helpers/3p/markdown.js
```diff
@@ -1,26 +1,26 @@
...
-});
\ No newline at end of file
+// const Code = require('code');
+// const Lab = require('lab');
+// const lab = exports.lab = Lab.script();
+// const expect = Code.expect;
+// const it = lab.it;
+// const describe = lab.describe
+//
+// const { buildRenderer } = require('../../spec-helpers');
```
jordanarldt, line 8: @rafa-avila-bc if we are commenting out this whole file, is there a reason we still need it?

helpers/3p/utils/lib/markdown.js
```diff
@@ -1,41 +1,53 @@
...
-    var ctx = merge({}, options, (this.context || this), context);
-    return md.render(options.fn(ctx));
-  }
+    function buildRemarkable(options) {
+        let remarkable;
+        if (options.linkify === true) {
+            remarkable = new Remarkable(options).use(linkify);
+        } else {
+            remarkable = new Remarkable(options);
+        }
```
jordanarldt, line 46:
```suggestion
        const remarkable = options.linkify === true
            ? new Remarkable(options).use(linkify)
            : new Remarkable(options);
```

This might look cleaner if we use ternary operators here, and then we don't need a mutable variable

spec/helpers/3p/markdown.js
rafa-avila-bc: I missed this. I already fixed it. Tests are now passing.

helpers/3p/utils/lib/markdown.js
rafa-avila-bc: done

jordanarldt (approved):

jmwiese (approved):

jairo-bc (approved):

PR #298, opened 2023-12-29
MERC-9364 Use CDN Original images for webdav - cache control

christensenep:
- [] [Paper Handlebars](https://github.com/bigcommerce/paper-handlebars/pull/273)
- [] [Paper](https://github.com/bigcommerce/paper/pull/343) (bump version)
- [ ] Stencil Cli (bump version)
- [ ] Storefront Renderer (bump verion) 

## What? Why?
Use CDN Original images for webdav - cache control. Take 2. See https://github.com/bigcommerce/paper/pull/343

If a developer/merchant decide to use the handlebar to import images from WebDav into the theme `({{cdn "webdav:/img/image.jpg"}})`, there is no cache-control header when the image is presented to the storefront.

Similar to our Image Manager, we should be using the cdn original images (Similar to Jinsoo's PR here [MERC-7127](https://github.com/bigcommerce/bigcommerce/pull/47254)

This PR differs from the previous as it only does this for file suffixes that image manager supports (jpg/jpeg/png/gif). Think people can get around that using webdav so I can expand the list if desired.

## How was it tested?
Tested locally and with unit tests


helpers/lib/cdnify.js
```diff
@@ -56,7 +56,10 @@ module.exports = globals => {
             }
 
             if (protocol === 'webdav:') {
-                return [cdnUrl, 'content', path].join('/');
+                const imgRegex = /(jpg|jpeg|gif|png)$/i;
+                const isImage = imgRegex.test(path);
+                const prefix = isImage ? 'images/stencil/original/content' : 'content'
+                return [cdnUrl, prefix, path].join('/');
```
jkanive, line 62: I feel like a logic as below is safer as it doesn't assume the image extension. What is your thought?
```js
if (path.length >= colonIndex + 5 && path[colonIndex+5] == '/') {
  const assetType = path.slice(colonIndex + 1, colonIndex + 5);
  let prefix = 'content'
  if (assetType == 'img/') {
    prefix = 'images/stencil/original/content';
  }
  return [cdnUrl, prefix, path].join('/');
} else {
   return [cdnUrl, 'content', path].join('/');
}
```

christensenep: I'm not aware that imagers need to be in the `img` path like that, if they are then I would agree. But just looking at my stores image manager images, they are not.

christensenep: If we can indeed rely on webdav images always being in that path then yeah that's definitely way better. (And perhaps most importantly, rely on ONLY images being there)

jkanive: There might be more types we might be missing. But we could make sure we cover all the supported in this https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types

christensenep: I was erring on the side of being conservative just given the history here, but yeah I may be overcompensating. Curious as to the opinion of storefront. 

Also realized I need to regex to match the '.', otherwise it'll match a file just named 'myjpg'.

christensenep:
♻️ I have tested this with `gif`s, `jpeg/jpg`s and `png`s, and verified that the new urls are valid and have cache control headers. I have also verified that other file extensions such as `pdf`s and, importantly, `bmp`s do *not* change the url from its current form and are thus still valid. I also verified that `bmp`s do *not* have the `images/stencil/content` form of the url available, so I believe this is the correct and exhaustive of extensions we should be doing this for.

christensenep:
@bigcommerce/team-merchandising @bigcommerce/team-storefront

jairo-bc (approved):

helpers/lib/cdnify.js
```diff
@@ -56,7 +56,10 @@ module.exports = globals => {
             }
 
             if (protocol === 'webdav:') {
-                return [cdnUrl, 'content', path].join('/');
+                const imgRegex = /.(jpg|jpeg|gif|png)$/i;
```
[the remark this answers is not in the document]
christensenep: Public docs are here: https://support.bigcommerce.com/s/article/Using-the-Image-Manager?language=en_US#requirements (referring to image manager, but it's the same list as webdav-uploaded images). The actual list in bcapp is https://github.com/bigcommerce/bigcommerce/blob/d1ea38e376fd8237a8c721b801ea7e6c36c3f87a/lib/Store/Controllers/Stencil/BaseImageController.php#L32

helpers/lib/cdnify.js
```diff
@@ -56,7 +56,10 @@ module.exports = globals => {
             }
 
             if (protocol === 'webdav:') {
-                return [cdnUrl, 'content', path].join('/');
+                const supportedImageExtensions = ['jpg', 'JPG', 'jpeg', 'JPEG', 'gif',  'GIF', 'png', 'PNG'];
+                const isImage = supportedImageExtensions.some(ext => path.includes(ext));
```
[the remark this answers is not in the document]
bc-alexsaiannyi: good catch, Jairo. Thanks!

jmwiese (approved):
LGTM

package.json
```diff
@@ -34,7 +34,7 @@
   "homepage": "https://github.com/bigcommerce/paper-handlebars",
   "dependencies": {
     "@bigcommerce/handlebars-v4": "4.7.8",
-    "chrono-node": "^2.6.5",
+    "chrono-node": "2.8.0",
```
bc-alexsaiannyi, line 37: FYI.  currently 2.8.1 is broken and temporary downgraded to 2.8.0

rtalvarez: i guess im also not following, why do we need the bump at all? your changes dont look to require this update at all

(and even in the entire repo, i can't find usages of this package)

bc-alexsaiannyi: in terms of experiment I only mean that we simply can turn it off, if something is broken and SFR would use original version of handlebars. 
 `package-lock` is under [gitignore](https://github.com/bigcommerce/paper-handlebars/blob/master/.gitignore#L3) and i don't see it in the project structure as well.

jairo-bc: @bc-alexsaiannyi feel free to merge master into your branch, this change is already [there](https://github.com/bigcommerce/paper-handlebars/pull/378)

jairo-bc (dismissed):

package.json
bc-alexsaiannyi: great, thanks!

helpers/lib/cdnify.js
```diff
@@ -56,7 +56,14 @@ module.exports = globals => {
             }
 
             if (protocol === 'webdav:') {
-                return [cdnUrl, 'content', path].join('/');
+                const supportedImageExtensions = ['jpg', 'jpeg', 'gif', 'png'];
+                const unsupportedPathSyntaxes = ['../', './'];
+                const searchParamIgnoredPath = path.split('?')[0];
```
bc-alexsaiannyi, line 61: we should cover cases when pathname contains searchParams like 
`
'webdav:/img/brands-logos/main.png?cache_bust=1749005114992'`
and exclude them from filtering

helpers/lib/cdnify.js
```diff
@@ -56,7 +56,14 @@ module.exports = globals => {
...
+                const isImage = supportedImageExtensions.some(ext => searchParamIgnoredPath.toLowerCase().endsWith(ext));
```
bc-alexsaiannyi, line 62: previously used method `includes()` can lead to incorrect results. For example, if  provided path looks like `'webdav:banners/menu-giftcard.svg'` then we will get situation where  `path.toLowerCase().includes(ext))` return `true` and helper modified path for unsupported content.

rtalvarez (commented):
nice, looks good 👍

helpers/lib/cdnify.js
```diff
@@ -56,7 +56,14 @@ module.exports = globals => {
             }
 
             if (protocol === 'webdav:') {
-                return [cdnUrl, 'content', path].join('/');
+                const supportedImageExtensions = ['jpg', 'jpeg', 'gif', 'png'];
```
jkanive, line 59: Should we include `.` ?

bc-alexsaiannyi:
jfyi, I would like to reiterate and summarise changes related to  webdav functionality: 

1. we currently support 4 types of images : `jpeg`, `jpg`, `png` and `gif`.
2. Content like `svg` or `pdf` files won't be affected.
3. We only provide cache-controlled url for images uploaded to `content` directory. That also means all previously uploaded content (images) to other directories won't be affected.

jairo-bc (approved):

PR #317, opened 2024-07-24
fix(storefront): STRF-12281 Prevent block and partial helpers from being named prototype methods

jordanarldt:
## What/Why?
Prevent `partial` and `block` helpers from allowing names that are the same as native javascript object prototype methods.

## Rollout/Rollback
Roll back this PR

## Testing
Unit tests, ensured that error is not thrown and object prototype is not accessed.

----

cc @bigcommerce/team-storefront 

[STRF-12281]: https://bigcommercecloud.atlassian.net/browse/STRF-12281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helpers/block.js
+++ b/helpers/block.js
@@ -10,6 +10,12 @@ const factory = globals => {
             globals.getLogger().info("Non-string passed to block helper");
             return '';
         }
+
+        if (Object.getOwnPropertyNames(Object.prototype).includes(name)) {
+            globals.getLogger().info(`Invalid name '${name}' passed to the partial helper. Returning empty string.`);
+            return '';
+        }
+
         const options = arguments[arguments.length - 1];
 
         /* Look for partial by name. */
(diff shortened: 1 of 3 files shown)


jmwiese (commented):
LGTM

huntario (approved):
💚
To test this out I used `npm link` to install this PR as a dependency of storefront-renderer-2 and then ran it locally. 

I added this to my home.html template file
```
<div class="container">
    {{#block "__proto__"}}STRF-12281 {{/block}}
</div>
```

This causes the store to throw an error
Screenshot 2024-07-24 at 11 54 58 AM

However, when I run STRF2 with this update, the server error no longer occurs and the error is handled
Screenshot 2024-07-24 at 11 54 23 AM

PR #324, opened 2024-08-19
feat: STRF-11900 Drop node 18, add node 20, 22

jairo-bc:
## What? Why?

- Breaking changes: Drop node support 16, 18
- Add node 20, 22
- Updated several packages related to CI
- Removed blackbird CI step

## How was it tested?

npm test

----


.nvmrc
```diff
@@ -0,0 +1 @@
+20
```
huntario, line 1: 🍹 
```
20

```

jordanarldt (approved):
Ran `npm install` and `npm test` on both node 20 and node 22, works as expected. Approved! 💚

PR #331, opened 2024-09-26
fix: STRF-12234 Make moment.js to warn instead of error

jairo-bc:
## What? Why?
- Make moment.js to warn instead of error

## How was it tested?
npm test
Screenshot 2024-09-25 at 18 25 32
Screenshot 2024-09-25 at 18 26 26

----

### DIFF
--- a/helpers/moment.js
+++ b/helpers/moment.js
@@ -43,6 +43,7 @@ const factory = () => {
             return moment(chrono.parseDate(str)).format(pattern);
         }
 
+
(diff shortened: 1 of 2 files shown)


bc-max (commented):
LGTM

ngaar (approved):
@jairo-bc personally, dont think this needs QA since you just changed error to warn. let me know if you think otherwise 🙇 

LGTM

PR #333, opened 2024-10-07
fix: STRF-12532 Set Access has been denied to resolve the property message to info level

jairo-bc:
## What? Why?
Set "Handlebars: Access has been denied to resolve the property" message to info level

## How was it tested?

`npm test`

----


bc-jz (commented):
LGTM

bc-jz (approved):
Tests cover this behavior and other systems need to pull in the new version of `paper-handlebars` for changes to occur. We can do QA on these other systems.

PR #339, opened 2025-02-07
Update math.js to include min, max, and clamp functions

34638a:
Include of simple math helpers for inline calculations. All support Number types of Int and Float, but not edge case numbers such as NaN, +-infinity, or undefined.

Min: given `A` and `B` return the smaller of the two on a standard number line.

Max: given `A` and `B` return the larger of the two on a standard number line.

Clamp: given `test`, `min` and `max` return the tested value if it falls within the range of `min` and `max` such that `min < test < max` is satisfied on a standard number line.  If the constraint is not satisfied, return the closest bounding constraint.

## What? Why?
Because these math functions are currently missing from the standard library.

## How was it tested?
None required, mirror of existing function `helper.add` with math operations swapped for logical operations.
Uses inline ternary operators for state evaluation. 

----

### DIFF
--- a/helpers/3p/math.js
+++ b/helpers/3p/math.js
@@ -135,3 +135,60 @@ helpers.avg = function() {
   args.pop();
   return exports.sum(args) / args.length;
 };
+
+/**
+ * Return the min of `a` and `b`.
+ * ```handlebars
+ * {{min 1 5}}
+ * //=> '1'
+ * ```
+ *
+ * @param {Number} `a`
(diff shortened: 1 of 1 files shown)


34638a:
Amended initial proposal notes with updated comments, and a testing protocol.

34638a:
Drafting due to issues with PR semantics as per previously outlined instructions by BC devs. Will correct and re-file when time permits.

PR #340, opened 2025-02-07
Create typeof.js

34638a:
Add implementation of javascript "typeof".
Implementation is based on json.js call which will yield a inline string as a result.

Registration into helpers list is required if approved. https://github.com/bigcommerce/paper-handlebars/blob/master/helpers.js

## What? Why?
Function to determine type of provided input regardless of what the inline result is for use in if statements, and inline object construction. eg. of using in JSON.Parse string `{"value_type":"{{typeof variable}}"}`.
This also has the side effect that it will prevent needed to serialise objects in no-js environments to determine flow state via CSS selectors.

```CSS
[data-test-for-typeof="Number"] {
    display: hidden;
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

## How was it tested?
Not tested. Based off existing file json.js with output line returning `typeof data` instead of `JSON.stringify(data)`.
----

### DIFF
--- a/helpers/typeof.js
+++ b/helpers/typeof.js
@@ -1,0 +1,14 @@
+'use strict';
+const common = require('./lib/common.js');
+
+const factory = globals => {
+    return function(data) {
+        data = common.unwrapIfSafeString(globals.handlebars, data);
+        return typeof data;
+    };
+};
+
+module.exports = [{
+    name: 'typeof',
+    factory: factory,
+}];


34638a:
@bc-evan-johnson
Hey mate,
I think this will be a good case for a learning opportunity as to what specifically BC would want for my pull requests when I submit them to make things as seamless as possible for your testing and integration. Because the purpose of this PR is to add a new tag to the system, can you please advise as to the documentation and testing protocol required to make adding a new helper like this a reality. Also this is not a complete PR based off the process required in #341 so it is not ready for review as at yet.

Looking forward to your response.
- Jordan

34638a:
> Hi @34638a!
> 
> I'm not really the main arbiter of such things - my primary role here is security review. To that end, I'm looking for code that
> 
> * serves a legitimate use-case (more code = more risk, so new code should add value)
> * is easy to understand & review
> * handles template inputs safely & does not introduce new weaknesses
> 
> Along with those points, backwards-compatibility is essential. Test coverage helps a lot with that :)
> 
> @jairo-bc and the rest of the team may have additional criteria, so I'll let them fill you in accordingly.

Thanks for your thoughts Evan,

Since writing the request for feedback / thoughts, I have pushed a second commit to this PR based off the requests made in the previous PR for the inline patch. Given your feedback, I'm pretty sure that this PR will meet all the criteria outlined above but I wont be going through the code again until Monday local time to confirm.

Have a good weekend,
- Jordan

helpers/typeof.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+const common = require('./lib/common.js');
+
+const factory = globals => {
+    return function(data) {
+        data = common.unwrapIfSafeString(globals.handlebars, data);
```
bc-evan-johnson, line 6: Helpers should always handle the Options object, otherwise there can be unexpected results. In this case, forgetting to pass an argument will lead to a return value of `object` rather than an error during development.
```suggestion
      return function(data) {
            arguments.pop(); // remove Options
            if (arguments.length !== 1) {
                throw new ValidationError("Handlebars helper 'typeof' requires exactly one argument.");
            }
            data = common.unwrapIfSafeString(globals.handlebars, data);
```

34638a:
Good shout on the suggested change.
I honestly didn't think to consider that someone may use the helper as a block when it's not meant to be and catch for it.

I'm surprised that the tests I wrote didn't catch for that even though I didn't explicitly test what happens when rendered as a block helper. I'll write a test later today that checks for if block, throw error.
@bc-evan-johnson

jairo-bc:
Hi @34638a, can you explain a bit more the use case over here? Do you think that some stencil context data type might be unstable (change per request), so you need to built a condition upon it? 
We are using semantic commit messages for PR titles and commits. CI Build error message should give a hint how to compose them

34638a:
Hi @jairo-bc 

Just responding to your message as timely as possible, corrections as per request will have to wait a day or so.

> can you explain a bit more the use case over here?

Regarding the usecase, there were two that came to mind at the time of writing as I came across a usecase in a theme I was editing:
1. In essence, a proxy function that is used in place of a long semantic chain `{{#if val1 "typeof" "string"}}string{{else if val1 "typeof" "number"}}number{{/if}} etc`  for copying data types out of "handlebars coding land" into HTML / CSS only semantics for use when a theme is intending to achieve goals without relying on JavaScript. For example, using the CSS data attribute selector `[data-type="number"]` one could use this in conjunction with `<div data-type="{{typeof 10}}">` to target specific elements that have numbers associated with them in handlebars (for example only, actual usecases would be for things like parsing arrays of unstructured data, like product customisation fields).
2. For use in a chained inline operation to avoid storing and reading temp variables over and over again. Using the newly accepted JSONparseSafe inline variant, one could construct a JSON object inline using the typeof operator `{{JSONparseSafe (multiconcat '{"val1_type":"' (typeof input_variable) '"}')}}`.


> Do you think that some stencil context data type might be unstable (change per request), so you need to built a condition upon it?

As per usecase reason 2. I can't remember the sepcifics but it came up in the rendering logic when attempting to assign the correct form input type to somewhere a customer could input data. The work around was to create several partials and a giant if block to do the same thing that could have been accomplished with a single type check, inflating the code and adding places where the duplicate code could be easily missed during refactor / edits (found out the hard way on this point).

Hope this provides insights as to the mindset that produced this PR.
- Jordan

34638a:
> We are using semantic commit messages for PR titles and commits.

Noted, didn't realise. Amended PR title to reflect intention. I'll go through and update the PR #350 also.

bc-evan-johnson:
> I honestly didn't think to consider that someone may use the helper as a block when it's not meant to be and catch for it.

I wasn't thinking of block usage either :)

The Options object is passed as the last argument to a Helper regardless of block/non-block usage, so if no args are passed in the template the Options object will be the first and only argument passed to the Helper function. I can't think of a good reason to be inspecting it directly in the template context, so if Options appears where a user-provided argument is expected that's probably a template bug (too few arguments).

34638a:
@bc-evan-johnson @jairo-bc Thoughts?
Given your past and above feedback, I think this is a winner thus far. I'll be testing by modifying a local copy of cornerstone NPM packages with the patch to test. I don't expect to run into any particular problems though. If this ends up working, I'll start on fixing up pr #350 more based off your guidance.

jairo-bc (approved):

34638a:
@jairo-bc before this gets merged, question: If I was feeling so inclined to author the docs entry for the `typeof` operator (and others) where should I look / submit PR's to? Some of the examples are kind of lacking in the current dev docs and it's actually annoying not having a stable reference in the docs of how some operators work. I find myself more often than not having to look for examples within cornerstone, or experimenting until it kind of works.

jairo-bc:
@34638a I think custom helpers documentation is available [here](https://github.com/bigcommerce/docs/blob/6c3f4c1ab0d92ca730c892914ff24429ce892b85/docs/storefront/stencil/themes/context/handlebars-reference.mdx), which will be renderer to https://developer.bigcommerce.com/docs/storefront/stencil/themes/context/handlebars-reference

34638a:
@jairo-bc Cheers, cloning the repo now, I'll open a PR there once done. Any particular person that should be tagged to let them know when it's done so it doesn't get merged in until after this PR is merged into main? I'll author a demo for the inline `JSONparseSafe` function too in a separate PR.
- Jordan

34638a:
@jairo-bc 

Thanks for assisting with the docs pr. I have a branch local I'll push up another savo with the basic examples for `typeof` operators and a lot of examples on how to use the json-safe as inline and in block form. Over time.

Next pr I'll clean up over here will likely be for the assignvar operation based off our discussions above.

- Jordan

PR #341, opened 2025-02-07
feat: Update jsonParseSafe.js - Inline use patch

34638a:
Add patch to allow for #JSONparseSafe to be used as a inline helper.

Examples:
`{{log (JSONparseSafe '{"test":"variable"}')}}`
`{{log (lookup (JSONparseSafe '{"test":{"validate":"variable"}}') "test")}}`

## What? Why?
To prevent this situation which makes code so much harder to maintain and reason with:
```Handlebars
{{! Pretend at this layer there is a pagination object here }}

{{! Before}}
{{#JSONparseSafe '{"test": "hello world!"}'}}
    {{#with this}}
        {{log test}}
        {{log ../../pagination}}
    {{/with}}
{{/JSONparseSafe}}


{{! After}}
{{#with (JSONparseSafe '{"test": "hello world!"}')}}
    {{log test}}
    {{log ../pagination}}
{{/with}}
```

This change removes a extra layer of context shifting `( ../ )` to access information about the context window.

## How was it tested?
Local tests, modified local theme testing file.
----

### DIFF
--- a/helpers/jsonParseSafe.js
+++ b/helpers/jsonParseSafe.js
@@ -5,7 +5,10 @@ function factory(globals) {
     const options = arguments[arguments.length - 1];
 
     try {
-      return options.fn(JSON.parse(value));
+      if (options.fn) {
+        return options.fn(JSON.parse(value));
+      }
+      return JSON.parse(value);
     } catch (err) {
       return options.inverse(this);
     }
--- a/helpers/jsonParseSafe.js
+++ b/helpers/jsonParseSafe.js
@@ -13,7 +13,7 @@ function factory(globals) {
       if (options.fn) { // If function block {{#JSONparseSafe "{}"}} escape here on excaption.
         return options.inverse(this);
       }
-      return {}; // Escape hatch to avoid crash due to inline malformed JSON input.
+      return undefined; // Escape hatch to avoid crash due to inline malformed JSON input.
     }
   };
 }
--- a/helpers/jsonParseSafe.js
+++ b/helpers/jsonParseSafe.js
@@ -10,7 +10,10 @@ function factory(globals) {
       }
       return JSON.parse(value);
     } catch (err) {
-      return options.inverse(this);
+      if (options.fn) { // If function block {{#JSONparseSafe "{}"}} escape here on excaption.
+        return options.inverse(this);
+      }
+      return {}; // Escape hatch to avoid crash due to inline malformed JSON input.
     }
   };
 }
--- a/spec/helpers/jsonParseSafe.js
+++ b/spec/helpers/jsonParseSafe.js
@@ -9,7 +9,7 @@ describe('JSONParseSafe helper', () => {
     };
     const runTestCases = testRunner({context});
 
-    it('should execute the main instruction', done => {
+    it('#Block: should execute the main instruction', done => {
         runTestCases([
             {
                 input: '{{#JSONparseSafe jsonString}}{{name}}{{/JSONparseSafe}}',
@@ -18,7 +18,7 @@ describe('JSONParseSafe helper', () => {
         ], done);
     });
 
-    it('should skip the main instruction if variable is non-json', done => {
+    it('#Block: should skip the main instruction if variable is non-json', done => {
         runTestCases([
             {
                 input: '{{#JSONparseSafe string}}{{name}}{{/JSONparseSafe}}',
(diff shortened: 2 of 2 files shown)


bc-evan-johnson (approved):

34638a:
@bc-evan-johnson
Hi Evan,
Forgot to test and escape for invalid input inline.
Added new commit to check state in catch block for if is or isn't a function and escape with a empty object if parsing fails.
New testing protocol below:
```Handlebars
{{! Pretend at this layer there is a pagination object here }}

{{! Before}}
{{#JSONparseSafe '{"test": "hello world!"}'}}
    {{#with this}}
        {{log test}}
        {{log ../../pagination}}
    {{/with}}
{{/JSONparseSafe}}


{{! After}}
{{#with (JSONparseSafe '{"test": "hello world!"}')}}
    {{log test}}
    {{log ../pagination}}
{{/with}}

{{! Tests for escape hatch inline to prevent crash on invalid json input }}
{{#with (JSONparseSafe '{')}} {{! single opening { as json input }}
    {{log test}} {{! empty / undefined }}
    {{log this}} {{! object '{}' }}
    {{log ../pagination}}
{{/with}}
```

helpers/jsonParseSafe.js
```diff
@@ -10,7 +10,10 @@ function factory(globals) {
       }
       return JSON.parse(value);
     } catch (err) {
-      return options.inverse(this);
+      if (options.fn) { // If function block {{#JSONparseSafe "{}"}} escape here on excaption.
+        return options.inverse(this);
+      }
+      return {}; // Escape hatch to avoid crash due to inline malformed JSON input.
```
bc-evan-johnson, line 16:
```suggestion
      return undefined; // Escape hatch to avoid crash due to inline malformed JSON input.
```
It's probably best to return `undefined` when no valid object or value could be parsed.

34638a: At this moment I don't have any outright objections to the response being `undefined` but I think it should be noted for future edits of the inline spec, or new helper developments.

The current block spec is as follows:
```handlebars
{{log "--------------"}}
{{#JSONparseSafe "{"}}
    {{log "safe block"}}
    {{log this}} {{!-- this is undefined if used as inline --}}
{{else}}
    {{log "else block"}}
{{/JSONparseSafe}}
{{log "--------------"}}
```

In the above example, `JSONparseSafe` will enter the false block.
Therefore I would expect that the following is valid code for use when calling as a inline:

```handlebars
{{#with (extend (JSONparseSafe 'BAD JSON') (JSONparseSafe '{"test":true}'))}}
    {{log this}} {{!-- logs { test: true } with current & original proposal --}}
{{/with}}
```

bc-evan-johnson: `(extend (JSONparseSafe 'BAD JSON') (JSONparseSafe '{"test":true}'))` should work regardless, since `extend` creates its own result object and skips non-object arguments (https://github.com/bigcommerce/paper-handlebars/blob/4fbc7b7efd6f6904753daa376a7c08a6d5055a0d/helpers/3p/object.js#L32-L37).

34638a: true, I think this is ready to merge / argue about internally at BC.
Good luck getting this through to production 😄  
- Jordan

34638a:
Just putting a bump on this so it doesn't sit here for a few years as a open PR.

@bc-evan-johnson :)
Can you please advise if there is a schedule for these kind of PR's to be reviewed so I can set expectations for submitted patches.

bc-evan-johnson (approved):

34638a:
Thanks @bc-evan-johnson I'll move onto the JSONparse function for inlineing next in that case.

34638a:
@bc-evan-johnson Any ideas on the turn around times for jairo to be able to review this? Asking because I may have accidently pushed a theme to production on our account that had this patch included :) It was a fun surprise that I forgot about. No rush, just curious.

jairo-bc:
@34638a looks good! Do you mind adding test cases (spec/helpers/jsonParseSafe.js) covering the scenarios described?

34638a:
@jairo-bc can do, I'll prepare them soon. It'll take a day or so to write cases that cover some expected inline uses and their outcomes based off the current spec.

jairo-bc (approved):

PR #347, opened 2025-03-14
fix: STRF-13111 Helpers last arguments improvements and fixes

jairo-bc:
## What? Why?

- Use last argument from arguments object instead of function arguments
- Use Object.create(null) instead of {}
- Add knownHelpers attribute

## How was it tested?

npm test

----

### DIFF
--- a/helpers/3p/collection.js
+++ b/helpers/3p/collection.js
@@ -19,16 +19,15 @@ var helpers = module.exports;
  *
  * @name .isEmpty
  * @param {Object} `collection`
- * @param {Object} `options`
  * @return {String}
  * @block
  * @api public
  */
 
-helpers.isEmpty = function(collection, options) {
-  if (!options) {
-    options = collection;
-    return options.fn(this);
+helpers.isEmpty = function(collection) {
+  const options = arguments[arguments.length - 1];
+  if (utils.isOptions(collection)) {
+    return collection.fn(this);
   }
 
   if (Array.isArray(collection) && !collection.length) {
@@ -47,13 +46,13 @@ helpers.isEmpty = function(collection, options) {
  *
  * @name .iterate
  * @param {Object|Array} `context` The collection to iterate over
- * @param {Object} `options`
  * @return {String}
  * @block
  * @api public
  */
 
-helpers.iterate = function(context, options) {
+helpers.iterate = function(context) {
+  const options = arguments[arguments.length - 1];
   if (Array.isArray(context)) {
     return forEach.apply(forEach, arguments);
   } else if (utils.isObject(context)) {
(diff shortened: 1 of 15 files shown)


index.js
```diff
@@ -76,6 +78,7 @@ class HandlebarsRenderer {
         for (let i = 0; i < helpers.length; i++) {
             const spec = helpers[i];
             this.handlebars.registerHelper(spec.name, spec.factory(this.helperContext));
+            handlebarsOptions.knownHelpers[spec.name] = true;
```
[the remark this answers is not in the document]
jairo-bc: not true
Screenshot 2025-03-17 at 12 10 28

helpers/3p/utils/index.js
```diff
@@ -373,14 +373,14 @@ utils.context = function (thisArg, locals, options) {
  * @api public
  */
 
-utils.options = function (thisArg, locals, options) {
+utils.options = function (thisArg, locals) {
```
bc-evan-johnson, line 376: This function isn't a helper, and is only used [here](https://github.com/bigcommerce/paper-handlebars/blob/260f8d1e67ecb6e9c38ef468c0a1e48fc81642f4/helpers/option.js#L24) where the helper implementation ensures the correct object is passed in. It probably doesn't need this change.

helpers/3p/utils/index.js
```diff
@@ -331,7 +331,8 @@ utils.isArray = Array.isArray;
  * @api public
  */
 
-utils.context = function (thisArg, locals, options) {
+utils.context = function (thisArg, locals) {
```
bc-evan-johnson, line 334: This is only used [here](https://github.com/bigcommerce/paper-handlebars/blob/260f8d1e67ecb6e9c38ef468c0a1e48fc81642f4/helpers/moment.js#L30) and the `moment` helper passed the correct options object. Since this isn't exposed as a Handlebars helper, the arguments change isn't necessary.

huntario (approved):
LGTM

helpers/3p/collection.js
```diff
@@ -19,16 +19,15 @@ var helpers = module.exports;
...
  */
 
-helpers.isEmpty = function(collection, options) {
-  if (!options) {
-    options = collection;
-    return options.fn(this);
+helpers.isEmpty = function(collection) {
+  const options = arguments[arguments.length - 1];
+  if (options === collection) {
+    return collection.fn(this);
```
bc-evan-johnson, line 30: It would be more clear what's going on here with something like
```suggestion
  if (arguments.length <= 1) {
    return options.fn(this);
```
or
```suggestion
  if (utils.isOptions(collection)) {
    return options.fn(this);
```

huntario (approved):

Issue #139, opened 2021-06-10
assignVar helper does not work with SafeString


peterrakmanyi:
1. All the helpers with a string output actually return a SafeString object. 
2. Not all helpers that accept a string input work with SafeString.
- custom helper: `{{assignVar "foo" (concat "b" "ar")}}` results in error:
`assignVar helper value must be a string or a number (integer)`

- built-in helper: `{{trim (concat "b" "ar")}}` fails to default `""` empty string

3. There is no helper to get a string from SafeString.

peterrakmanyi:
It appears that there is a workaround with using the `get` helper like this:
`{{get 'string' (concat "b" "ar")}}` results in `"bar"`

...

`{{assignVar "foo" (get 'string' (concat "b" "ar"))}}` does result in `var foo="bar"`

Issue #305, opened 2024-05-01
Loosen AssignVar Limits


bobbyshaw:
Hi,

Our team at Space 48 recently had a production issue where a customer adding a widget to a global region on a page in PageBuilder broke content rendering on other pages. The cause was the use of assignVar in this widget took them over the 50 vars limit on these pages.

The out-of-hours incident was attended to by our team and supported by BigCommerce support team, however both were unable to immediately identify that this was the cause. 

Upon review, the use of assignVar in this project seems reasonable. It has been used in numerous places to simplify the template logic by pulling business logic to the top of the template and using variables as flags or the result of if/else check. 

Unfortunately, given the impact and lack of traceability when it occurs we have since discouraged developers from using assignVar and refactored our project to reduce the use of it.

Moving forward, we would like to:

1. Be able to more easily identify when this is the cause of a storefront outage. 
2. Be able to monitor assignVar use per page
3. Encourage assignVar use if it helps reduce the complexity of the template. 

### Identifying this issue

I'm afraid I don't have any ideas here so I'm open to suggestions or education if there's already a log somewhere of stencil issues in production! In our instance, the issue wasn't immediately repeatable by running a copy of the theme locally against production. 

### Monitoring assignVar use

Our understanding is that the current limit is 50 per page/template. Our challenge is that beyond a check for assignVar use across the entire theme, we don't have a reliable way of monitoring assignVar use per template, especially when you factor in widget placements as well. 

Is there or could there be a way for us to monitor the amount of vars on a page? For example, a new helper could return the count of variables. We could use something like this in the footer to keep an eye on real-world usage. 

```
Object.keys(globals.storage.variables).length
```

### Encourage assignVar use

I understand that the limit of 50 was put in place as a safety mechanism. While this is the first time that we've come across it, I wonder if there is any appetite to increase this limit substantially so that it can be used without fear while also still serving as protection for memory usage. 

Tagging for visibility: @bookernath @AndrewBarber

### Appendix

For your reference, here's a list of 46 assignVars that I've identified in our theme around the time of the incident to give you a flavour of what they were being used for. https://gist.github.com/bobbyshaw/88b1041d791cb9de9aa15ff575611344

AndrewBarber:
Tickets with BC support for this was 07721161 (may help?). I believe there was a subsequent ticket, however, failing to find it on mobile. 😅

Tiggerito:
[...]

PR #350, opened 2025-03-30
feat(assignVar): Patch for assignVar to add "drop key" feature and fixes reflecting documentation

34638a:
## Proposal, What? Why?

Helper `assignVar` currently does not have a way to do any of the following features:
1. Un-Map a assigned value: once a key has been set, it is locked for the whole render pipeline. This results in features like widgets taking up half of the available variables only to use them once and never again, limiting theme developers from being able to add new features due to limited variable space.
2. Accept an empty string into a variable as its value. Given a string is valid input, all states of a string should be considered as valid, limited by length. For instance, if a variable is set to `""` because it is intended to be a placeholder when a value is not provided by the store admins or a customer input, say for a phone number click to call, this will currently cause the store to err.500 instead of handling what should be a valid input. The data input is inherently reasonable, so there is a patch in the validation logic to accept an empty string value instead of rejecting it.
3. Ingest a `SafeString` object and treat it like a actual string for purposes of storage and retreival. Current workaround possible as per open issue #139

## Changes Overview
Create patch for assignVar to add following features:
- Empty string ( "" ) to be accepted as entry into assignVar helper. 
- Null or Undefined values as input to assignVar now delete the key that was set. This has the use-case of freeing variables assigned by widgets or temporary assignments.
- Internal conversion via `unwrapSafeString` call before variable is process to all for storage and retrieval of `SafeString` contents.
- Tests for these use-cases in assignVar-getVar.js

## How was it tested?
Test cases provided in `assignVar-getVar.js` file.

----

### DIFF
--- a/helpers/assignVar.js
+++ b/helpers/assignVar.js
@@ -7,12 +7,12 @@ const common = require("./lib/common");
 
 const factory = globals => {
     return function (key, value) {
+        globals.getLogger().info(`DEBUG assignVar helper called with key: ${key} and value: ${value}`);
 
         // Validate that key is a string
         if (!utils.isString(key)) {
             throw new ValidationError("assignVar helper key must be a string");
         }
-
 
         // Setup storage
         if (typeof globals.storage.variables === 'undefined') {
@@ -30,7 +30,7 @@ const factory = globals => {
                 throw new ValidationError("assignVar helper value must be a string or a number (integer/float)");
             }
 
-            // Validate that string is not longer than the max length
+            // Validate that string is not longer than or equal to the max length
             if (utils.isString(value) && value.length >= max_length) {
                 throw new ValidationError(`assignVar helper value must be less than ${max_length} characters, 
                 but a ${value.length} character value was set to ${key}`);
@@ -53,4 +53,6 @@ const factory = globals => {
 module.exports = [{
     name: 'assignVar',
     factory: factory,
+    max_length: max_length,
+    max_keys: max_keys
 }];
--- a/helpers/assignVar.js
+++ b/helpers/assignVar.js
@@ -12,29 +12,36 @@ const factory = globals => {
             throw new ValidationError("assignVar helper key must be a string");
         }
 
-        // Validate that value is a string or Number (int/float)
-        if (!utils.isString(value) && !Number.isFinite(value)) {
-            throw new ValidationError("assignVar helper value must be a string or a number (integer/float)");
-        }
+        //Check for if the assigned value is being set to null or undefined
+        if (!(value === null || value === undefined)) {
 
-        // Validate that string is not longer than the max length
-        if (utils.isString(value) && value.length >= max_length) {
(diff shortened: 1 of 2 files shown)


34638a:
Reference to open issue #305

34638a:
Reference to open issue #139
^ This will need to be added as a part of the "string" type checks / validation logic.
Note for @34638a to add this into the next PR push for proposal.

34638a:
Rename title of PR due to 100 character limit: 
~~feat(assignVar): Patch for assignVar to add feature and amend function reflecting documentation / scope.~~ feat(assignVar): Patch for assignVar to add "drop key" feature and fixes reflecting documentation.

34638a:
Hey @jairo-bc 

Clarification request: given the line comment in the actual code, this reads as incorrect implementation. Can you please confirm if the below code should be interpreted as "the buffer length of 1024 should contain no more than 1024", or as per the current implementation "the buffer cannot hold 1024 character, or more than 1024 characters. The limit is 1023 characters in the buffer."

``` Javascript
        // Validate that string is not longer than the max length
        if (utils.isString(value) && value.length >= max_length) {
            throw new ValidationError(`assignVar helper value must be less than ${max_length} characters, 
                but a ${value.length} character value was set to ${key}`);
        }
```

- Jordan

34638a:
@jairo-bc @bc-evan-johnson 
Other than the above @ for clarification,
I think it's ready for review. Thoughts and feedback are appreciated.
- Jordan

helpers/assignVar.js
```diff
@@ -3,42 +3,57 @@ const utils = require('./3p/utils');
...
-        // Store value for later use by getVar helper
-        globals.storage.variables[key] = value;
+            // Validate that string is not longer than or equal to the max length
+            if (utils.isString(value) && value.length >= max_length) {
+                throw new ValidationError(`assignVar helper value must be less than ${max_length} characters, 
+                but a ${value.length} character value was set to ${key}`);
+            }
+
+            // Make sure the number of total keys is within the limit
+            if (Object.keys(globals.storage.variables).length >= max_keys) {
```
jairo-bc, line 40: You are right! Looks like the limit is 49 not 50 :(

helpers/assignVar.js
```diff
@@ -3,42 +3,57 @@ const utils = require('./3p/utils');
 const max_length = 1024;
 const max_keys = 50;
 const { ValidationError } = require('../lib/errors');
+const common = require("./lib/common");
 
 const factory = globals => {
     return function (key, value) {
+        //globals.getLogger().info(`DEBUG assignVar helper called with key: ${key} and value: ${value}`);
```
jairo-bc, line 10: Do you mind removing debugging line?

34638a: Give me 15, I'll update the PR 😎

helpers/assignVar.js
34638a: Edited code to adjust to "inclusive of limit" when assigning variables :)

jairo-bc (approved):

PR #354, opened 2025-04-09
fix: STRF-13111 Add stack trace to RenderError; arguments improvements for array helpers

jairo-bc:
## What? Why?

- Added stack trace to RenderError message, that is used for rendering scripts
- arguments improvements for array helpers

## How was it tested?

npm test

----

### DIFF
--- a/helpers/3p/array.js
+++ b/helpers/3p/array.js
@@ -77,15 +77,15 @@ helpers.before = function(array, n) {
  * {{/eachIndex}}
  * ```
  * @param {Array} `array`
- * @param {Object} `options`
  * @return {String}
  * @block
  * @api public
  */
 
-helpers.eachIndex = function(array, options) {
+helpers.eachIndex = function(array) {
   var result = '';
   if (Array.isArray(array)) {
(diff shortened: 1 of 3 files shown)


jmwiese (dismissed):

jmwiese:
@jairo-bc any chance this breaks some themes/scripts if they're using helpers that will begin to be disallowed that werent previously?

jairo-bc:
@jmwiese yes, it will throw an error that you are using non-registered helper. Right now, it would just swallow and render empty string.
It was already part of the previous attempt and in a day I saw around 50 instances. Those were mostly misspelling helpers name.
Do you think it's a breaking change and should be comminucated?

jmwiese:
@jairo-bc I think we might want to let like ninja and stuff know. as long as it doesnt cause the page to just 500 and instead would just be missing some content or something I think it's ok. are there any valid/functional helpers that people might use that you know of that would break because of this?

jairo-bc:
@jmwiese I'm going to remove that attribute.
- It will make a page 500 instead of just empty string. 
- For scripts, it will break the render of a single one completely

